PR #86, opened 2017-12-14
Rid of description fields in boolean and object to get rid of repeate…

### LINKED ISSUE #84: Booleans repeat the description
image

We need to override booleans so we aren't repeating data

uppal101:
…d description. Fix #84


uppal101:
I left the extra schemas in there for later since we addressed at some point array and object description repetition will need to be fixed

packages/api-explorer/__tests__/form-components/CheckboxWidget.test.jsx
```diff
@@ -0,0 +1,16 @@
...
+  const checkbox = shallow(
+    <Checkbox
+      id="id"
+      label="label"
+      required
+      onChange={() => {}}
+      schema={{ description: 'description' }}
+    />,
+  );
+  expect(checkbox.html()).not.toMatch(/description/);
```
uppal101: Question the .not.toMatch(/description/) means the string description shouldn't show up in the snapshot correct?

packages/api-explorer/src/form-components/ObjectField.jsx
```diff
@@ -55,13 +55,6 @@ class CustomObjectField extends ObjectField {
             formContext={formContext}
           />
         )}
-        {(uiSchema['ui:description'] || schema.description) && (
```
uppal101: This change came from the original directory I had locally I think we can keep it as we discussed until we figure out how to design readme to be recursive?

example/swagger-files/oas.json
```diff
@@ -0,0 +1,274 @@
+{
```
uppal101: should this file be deleted?

packages/api-explorer/__tests__/form-components/CheckboxWidget.test.jsx
domharrington: I removed the snapshot test. It was simple enough to do it without snapshots

packages/api-explorer/src/form-components/ObjectField.jsx
domharrington: Ahhh yeah i think we should revert this change until we've fully figured out what the repercussions are and fixed that issue separarely?

Issue #101, opened 2018-02-12
Add in a way to bring in different codemirror modes


domharrington:
It's gunna be unstyled if we try and bring in a different language than those listed here: https://github.com/readmeio/api-explorer/blob/9b4b01c76d9b44f4c70d99abc06c16c8ece0226b/packages/syntax-highlighter/codemirror.js#L2-L8

Maybe we can use some form of dynamic imports to bring in some languages that aren't currently supported: https://webpack.js.org/guides/code-splitting/#dynamic-imports

PR #102, opened 2018-02-12
Bug/fix syntax highlighting

domharrington:
There was a bug in the syntax highlighter which meant it couldn't syntax highlight java and HTML. I've fixed that and also made sure the Examples are correctly syntax highlighting.

### DIFF
--- a/example/src/Demo.jsx
+++ b/example/src/Demo.jsx
@@ -67,7 +67,7 @@ class Demo extends React.Component {
               }}
               flags={{ correctnewlines: false }}
               suggestedEdits
-              oauthUrl={this.props.oauthUrl}
+              oauth={this.props.oauth}
             />
           )
         }
@@ -77,10 +77,10 @@ class Demo extends React.Component {
 }
 
 Demo.propTypes = {
-  oauthUrl: PropTypes.string,
+  oauth: PropTypes.bool,
 };
 
 Demo.defaultProps = {
-  oauthUrl: '',
+  oauth: false,
 };
 module.exports = Demo;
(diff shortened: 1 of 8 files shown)


example/src/Demo.jsx
```diff
@@ -67,7 +67,7 @@ class Demo extends React.Component {
               }}
               flags={{ correctnewlines: false }}
               suggestedEdits
-              oauthUrl={this.props.oauthUrl}
+              oauth={this.props.oauth}
```
uppal101: This isn't a change from the new API key, is it?

packages/syntax-highlighter/codemirror.js
```diff
@@ -6,14 +6,73 @@ require('codemirror/mode/shell/shell');
...
+  //   if (info.mime == lang) {
+  //     modeName = info.mode;
+  //   } else if (info.name.toLowerCase() === lang) {
+  //     modeName = info.mode;
+  //     lang = info.mime;
+  //   }
+  // });
   // if (!CodeMirror.modes[lang]) {
   //   require(`codemirror/mode/${lang}/${lang}.js`);
   // }
```
uppal101: Do we need this if block it seems to be commented out from last time?

example/src/Demo.jsx
domharrington: I recently renamed `oauthUrl` to `oauth` in the ApiExplorer props, but this was still using the old way.

packages/syntax-highlighter/codemirror.js
domharrington: There's a TODO at the top of the block linking to this: https://github.com/readmeio/api-explorer/issues/101 which we will eventually need to do.

Issue #29, opened 2017-09-13
Need to figure out what to do about angular links in markdown


domharrington:
https://github.com/readmeio/api-explorer/blob/9639c0742a8082328eadd63459b1c373476c11ed/packages/api-explorer-ui/src/lib/marked.js#L80

domharrington:
See the problem here (click this link "Getting Started with swagger-pet-store1"):

This is using the old api explorer:
https://swagger-pet-store1.readme.io/v1.0/reference#updatepet

It works because angular and pjax hijack the link and render the new page content.

This does not work:
https://swagger-pet-store1.readme.io/v1.0/reference?newApiExplorer=true#updatepet

I think we need to remove everything to do with uiSref from the markdown and just use normal hyperlinks in their place.

- [ ] https://github.com/readmeio/api-explorer/blob/9b4b01c76d9b44f4c70d99abc06c16c8ece0226b/packages/api-explorer/src/lib/markdown/renderer.js#L81 should be replaced with `/blog/:blog`
- [ ] https://github.com/readmeio/api-explorer/blob/9b4b01c76d9b44f4c70d99abc06c16c8ece0226b/packages/api-explorer/src/lib/markdown/renderer.js#L64 should be replaced with `/doc/:doc`
- [ ] https://github.com/readmeio/api-explorer/blob/9b4b01c76d9b44f4c70d99abc06c16c8ece0226b/packages/api-explorer/src/lib/markdown/renderer.js#L87 should be replaced with `/page/:custompage`

domharrington:
Right now our markdown renderer is outputting the following:

From this:
```
[Getting Started with swagger-pet-store1](doc:getting-started)
```

To this:

```
<a ui-sref="docs.show({'doc': 'getting-started'})">
```

We need to change it to a regular link, so it outputs:

```
<a href="/doc/getting-started">
```

uppal101:
#103

PR #103, opened 2018-02-14
Get rid of uiSref and use href instead

uppal101:


packages/api-explorer/src/lib/markdown/renderer.js
```diff
@@ -78,13 +77,12 @@ renderer.link = function link(href, title, text) {
 
   const blog = href.match(/^blog:([-_a-zA-Z0-9#]*)$/);
   if (blog) {
-    uiSref = `blog.show({'blog': '${blog[1]}'})`;
-    href = '';
+    href = `/blog/${blog[1]}`;
```
uppal101: I know you put in the issue to change it to '/blog/:blog', but how will it have access to the variable blog shouldn't we being use string interpolation?

domharrington: This line of code looks correct to me?

packages/api-explorer/__tests__/lib/markdown/renderer.test.js
```diff
@@ -0,0 +1,29 @@
...
+      '<a href="/v1.0/blog/getting-started" target="_self">Getting started with petstore swagger</a>',
+    );
+  });
+
+  it('should return a tag with href', () => {
+    expect(renderer.link('/v1.0/page/getting-started', '', '')).toBe(
+      '<a href="/v1.0/page/getting-started" target="_self"></a>',
+    );
+  });
+});
```
domharrington: You've not added any tests that have `doc:doc` as the href, to test that we're correctly converting those into /doc links.

Can you add one for each type: doc, blog and custompage?

packages/api-explorer/src/lib/markdown/renderer.js
```diff
@@ -53,16 +53,15 @@ renderer.link = function link(href, title, text) {
   /* eslint no-param-reassign: 0 */
   const doc = href.match(/^doc:([-_a-zA-Z0-9#]*)$/);
   let isDoc = false;
-  let uiSref = false;
+  const uiSref = false;
```
domharrington: We no longer need this now?

domharrington:
This fixes #29

PR #108, opened 2018-03-09
Feature/custom code samples

uppal101:
I know js-inspect is failing due to the new way codeSample is now being rendered in Doc. Also a test is failing on Example regarding syntax highlighting not 100% sure how to go forward with that one.


packages/api-explorer/src/CodeSample.jsx
```diff
@@ -5,61 +5,133 @@ const Oas = require('./lib/Oas');
...
-        const snippet = generateCodeSnippet(oas, operation, formData, language);
+class CodeSample extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      customCodeSampleTab: 0,
+    };
+    this.setCustomCodeSampleTab = this.setCustomCodeSampleTab.bind(this);
+  }
+  setCustomCodeSampleTab(index) {
```
domharrington: How come we need to do this? How come `setLanguage()` isnt enough?

domharrington:
It looks like the coverage has slipped because we're not testing `setCustomCodeSampleTab()`. Is it possible to test this?

packages/api-explorer/src/CodeSample.jsx
uppal101: It's so that the code samples body can switch along with  whatever language is set here https://github.com/readmeio/api-explorer/pull/108/files#diff-c199d2e44031baaac9109fd82d5610cdR54

I can try to make the styling work with setLanguage?

PR #109, opened 2018-03-13
Feature/css forms

uppal101:
There are still issues with the orderable buttons on hover you can still see the border, also they still take up space and can be selected.

Also there is the issue of functionality of the delete button which errors due to format int64


example/react-json-schema-form.css
```diff
@@ -0,0 +1,63 @@
...
+}
+.btn-add::after {
+  font-family: 'ReadMe-Icons';
+  content: "\e975";
+  color: #028aec;
+}
+.btn.btn-default {
+  border: none;
+}
+.array-item-move-up::after {
```
uppal101: The styling on the following lines until line 57 never get reflected but I'm sure at some point we will them to work. The buttons still take up space and are clickable .

PR #116, opened 2018-04-05
Use password input type for basic auth

rossover:
Updated *mixins.jade* to use an input type of `password` instead of `text` to not show the password on-screen when demoing to others, etc.


domharrington:
Hey! The file you edited was actually in the legacy folder, which is used for a reference implementation of what readme currently does. It's currently a feature that the password is visible, as it's often used for an API key, or a single use password, instead of an actual username/password combination.

Thanks for contributing though.

PR #125, opened 2018-06-29
Feature/jwt latest

domharrington:
- [x] check to see if variables work in code samples
- [x] need to make sure dropdowns are rendered into another element than inline in the markdown (maybe use Portals? https://reactjs.org/docs/portals.html). Right now the styling is kinda messed up because we're rendering the dropdown inline in the same element as the markdown container, so the list item styles are being inherited from there.

screen shot 2018-07-02 at 16 24 28

It should look like this:

screen shot 2018-07-02 at 16 25 35


domharrington:
The code changes start here in the diff: https://github.com/readmeio/api-explorer/pull/125/files#diff-40d0fdc6094816dce8d850ec98f0b83d

The stuff above it is emojis that i added for testing.

domharrington:
Also in this PR I changed the syntax-highlighter API, so we should probably do a major bump when this gets merged in!

domharrington:
I've also added @readme/markdown as a new package which is a function to convert markdown into an array of React components.

Right now the renderMagic equivalent is still stuck inside the api-explorer package, so I'm thinking maybe I should perform a further refactor to move https://github.com/readmeio/api-explorer/blob/master/packages/api-explorer/src/block-types/Content.jsx into @readme/markdown as well? Maybe as a separate export? 

Then we could potentially use this as a replacement for docs rendering on main readme.

packages/api-explorer/src/ResponseBody.jsx
```diff
@@ -9,13 +9,9 @@ function Authorized({ result }) {
...
       {result.isBinary && <div>A binary file was returned</div>}
       {!result.isBinary && (
-        <div
-          className="cm-s-tomorrow-night codemirror-highlight"
-          // eslint-disable-next-line react/no-danger
-          dangerouslySetInnerHTML={{
-            __html: syntaxHighlighter(JSON.stringify(result.responseBody), 'javascript'),
-          }}
-        />
+        <div className="cm-s-tomorrow-night codemirror-highlight">
```
mjcuva: Do you think this should be part of the component returned from the syntaxHighlighter?

domharrington:
Should revert back to using enzyme latest now that React.Context support is there: https://github.com/airbnb/enzyme/issues/1509

domharrington:
- [ ]  make sure this works with #130

packages/api-explorer/src/CodeSample.jsx
```diff
@@ -92,11 +90,7 @@ class CodeSample extends React.Component {
               </ul>
               <div className="hub-code-auto">
                 <CopyCode code={code} />
-                <pre
-                  className={`tomorrow-night hub-lang hub-lang-${language}`}
-                  // eslint-disable-next-line react/no-danger
-                  dangerouslySetInnerHTML={{ __html: snippet }}
-                />
+                <pre className={`tomorrow-night hub-lang hub-lang-${language}`}>{snippet}</pre>
```
mjcuva: Why does snippet no longer need to be set with `dangerouslySetInnerHTML`

packages/markdown/components/Anchor.jsx
```diff
@@ -0,0 +1,63 @@
+const React = require('react');
+const PropTypes = require('prop-types');
+
+// Nabbed from here:
+// https://github.com/readmeio/api-explorer/blob/0dedafcf71102feedaa4145040d3f57d79d95752/packages/api-explorer/src/lib/markdown/renderer.js#L52
+function getHref(href) {
+  const doc = href.match(/^doc:([-_a-zA-Z0-9#]*)$/);
+  if (doc) {
+    return `/docs/${doc[1]}`;
```
mjcuva: Does all of this stuff break for enterprise stuff? Ie when the url should be `developers.intercom.io/building-apps/docs`

packages/api-explorer/src/CodeSample.jsx
domharrington: The syntax-highlighter returns react elements now! Because we have to be able to swap out variables in the middle of code blocks: https://github.com/readmeio/api-explorer/blob/3147975c4c52d914ce7964faaa5e2ed0b6ddff0d/packages/syntax-highlighter/index.js

packages/markdown/components/Anchor.jsx
domharrington: That's a very good question... i've not changed this piece during the refactor, it came from the old markdown renderer: https://github.com/readmeio/api-explorer/blob/0dedafcf71102feedaa4145040d3f57d79d95752/packages/api-explorer/src/lib/markdown/renderer.js#L52

I'm not sure how this should be tackled really. Does this work in readme already? I wonder how we do it over there.

domharrington: Just tested this locally. The old explorer uses `ui-sref=docs.show({'doc': 'slug'})` which works to give the url `http://enterprise-group-parent.readme.local:3000/child/v1.0/docs/slug`. The new explorer gives a URL of `http://enterprise-group-parent.readme.local:3000/docs/slug`

domharrington: Wonder if we could give a basePath or something which it can use to construct all of these links?

mjcuva: Hmm we'll probably have to do something like that, since I can't think of another way the api explorer will be aware it's in a child project. Annoying to have to pass another thing through though. I think since this is currently broken we should make it an additional task and not let it hold up this pr

PR #130, opened 2018-08-02
Add a button to copy code samples

mjcuva:
Add a button to copy code samples!


packages/api-explorer/src/CodeSample.jsx
```diff
@@ -35,7 +37,8 @@ class CodeSample extends React.Component {
         </ul>
         <div className="code-sample-body">
           {examplesWithLanguages.map(example => {
-            return (
+            return [
+              <CopyCode code={example.code} />,
```
domharrington: Does this not make React complain about there not being a `key` on the array items returned?

packages/api-explorer/src/block-types/Code.jsx
```diff
@@ -57,7 +58,8 @@ class BlockCode extends React.Component {
           )}
 
           <div className="block-code-code">
-            {codes.map((code, i) => (
+            {codes.map((code, i) => [
+              <CopyCode code={code.code} />,
```
domharrington: I should've noticed this. Will this also have the same key problem?

PR #131, opened 2018-08-03
Add error handling around `await response.json()` call

domharrington:
This is to catch cases where they return `content-type: application/json` but do not return valid JSON

We have to clone the object first before we parse, as the objects can only be read once


domharrington:
Also added a default URL for if their OAS file has no servers array.

PR #132, opened 2018-08-03
Feature/restyling

domharrington:


domharrington:
Overall this is so good. Much better than what we currently have. 

- [ ] make types slightly lighter, whilst still looking distinct when there's a description

screen shot 2018-08-03 at 17 24 54

- [ ] can we do something better to booleans?

screen shot 2018-08-03 at 17 25 13

- [ ] can we make array descriptions look better somehow?

screen shot 2018-08-03 at 17 25 54

- [ ] is there anything we can do where some things have descriptions and others dont. The type makes it look slightly weird as it's floating against the description instead
screen shot 2018-08-03 at 17 27 15

domharrington:
Excellent work. 

- [ ] Can you try and make booleans to have the same layout as the other types?

screen shot 2018-08-07 at 10 59 57

I'd love the checkbox to be on the right, with the other inputs. Label far left, then type, then description in the middle. The markup appears to be different though, so may not be possible?

dok:
https://github.com/readmeio/readme/pull/1138/comment#issuecomment-411592145

PR #133, opened 2018-08-14
Feature/json type

domharrington:


packages/api-explorer/src/form-components/FileWidget.jsx
```diff
@@ -0,0 +1,14 @@
+const extensions = require('@readme/oas-extensions');
```
mjcuva: Isn't there already a way to upload a file in the new api explorer?

packages/api-explorer/src/lib/oas-to-har.js
```diff
@@ -156,15 +156,30 @@ module.exports = (
...
   if (schema.schema && Object.keys(schema.schema).length) {
     // If there is formData, then the type is application/x-www-form-urlencoded
     if (Object.keys(formData.formData).length) {
       har.postData.text = querystring.stringify(formData.formData);
     } else if (isPrimitive(formData.body) || Object.keys(formData.body).length) {
-      // Default to JSON.stringify
-      har.postData.text = JSON.stringify(
-        typeof formData.body.RAW_BODY !== 'undefined' ? formData.body.RAW_BODY : formData.body,
-      );
+      try {
```
mjcuva: Can you add some comments about what this is doing? It's not really obvious when it falls into this case

packages/api-explorer/src/form-components/FileWidget.jsx
domharrington: I added this wrapper so that when `x-explorer-enabled: false` then it doesn't show the <input>

PR #139, opened 2018-08-28
Feature/react error boundary

domharrington:
Example:
screen shot 2018-08-28 at 09 56 36

Example within readme:
Top level error boundary:
screen shot 2018-08-30 at 15 37 41

Endpoint error boundary:
screen shot 2018-08-30 at 15 38 12

I'm hoping that the only reason the stack traces are bad is due to React being in development.

In future i'd like for this to report to bugsnag or something, but I think this is okay for now.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -17,5 +17,5 @@ module.exports = merge(common, {
     new webpack.NamedModulesPlugin(),
     new webpack.HotModuleReplacementPlugin(),
   ],
-  devtool: 'eval-source-map',
+  devtool: 'cheap-module-source-map',
 });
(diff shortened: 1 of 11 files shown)


webpack.config.js
```diff
@@ -17,5 +17,5 @@ module.exports = merge(common, {
     new webpack.NamedModulesPlugin(),
     new webpack.HotModuleReplacementPlugin(),
   ],
-  devtool: 'eval-source-map',
+  devtool: 'cheap-module-source-map',
```
mjcuva: What's this change for?

domharrington: https://github.com/readmeio/api-explorer/commit/d412bc2680979f71b133f2e1978c6e6fc08a3307

PR #140, opened 2018-08-28
Fix regression introduced in 0.9.4 which caused the reference guides to fail

domharrington:


domharrington:
When you create a category and pages in the reference section, and not an API

screen shot 2018-08-28 at 11 57 47

Then there is no underlying apiSetting. apiSetting is only required if `type='endpoint'`

PR #151, opened 2018-09-24
Feature/oas preview

domharrington:
Add preview site to replace https://swagger.readme.io/preview/aHR0cDovL3BldHN0b3JlLnN3YWdnZXIuaW8vdjIvc3dhZ2dlci5qc29u#tag__store


mjcuva:
Shouldn't it be using the path instead of the tag as the page title?

domharrington:
I think you're right... though this also appears to be broken on here too: https://readmeio.github.io/api-explorer/?selected=https://readme.readme.io/openapi/5a95f7d15c3020002b02aace

I think this would be fixed by adding path to this line? https://github.com/readmeio/api-explorer/blob/94c3984367c492c03d27cfa62364ab688a48c457/packages/api-explorer/lib/create-docs.js#L44

domharrington:
Fixed the issue with the tags, now they get the path:

screen shot 2018-09-26 at 15 46 40

Also added some error handling if whatever they give us is invalid JSON etc:

screen shot 2018-09-26 at 15 47 42

We do not validate their swagger file here like we used to, but i think this is fine for now.

PR #160, opened 2018-10-05
fix: TypeError: e.children[0].toLowerCase in Heading.jsx (#159)

### LINKED ISSUE #159: TypeError: e.children[0].toLowerCase is not a function
This is an issue with our markdown parser. We have to construct an id based upon the text content of the header. So for the following:

```
# This is a header
```

We assign an id to that element of 

```
section-this-is-a-header
```

This breaks when there is a code highlighted element as the first part of a header:

```
# `header`
```

This happens here: https://github.com/readmeio/api-explorer/blob/d2df77e92fc0be9cdce2d7694f651265e15b34d9/packages/markdown/components/Heading.jsx#L5

Because `children[0]` is not text, but is a Code element that gets created here: https://github.com/readmeio/api-explorer/blob/d2df77e92fc0be9cdce2d7694f651265e15b34d9/packages/markdown/index.js#L42

This can be tested locally by uncommenting this block here: https://github.com/readmeio/api-explorer/blob/d2df77e92fc0be9cdce2d7694f651265e15b34d9/packages/api-explorer/lib/create-docs.js#L29 and adding: "# \`code\`" to the end of the body element.

This will result in an error like this:

screen shot 2018-10-04 at 17 51 51

nadalfederer:
closes #159


domharrington:
This looks great! Thanks.

PR #163, opened 2018-10-18
fix: allow to pass non-undefined values(#149)

### LINKED ISSUE #149: parameter bug, root cause: the value of boolean parameter is false
1. If we define "false" as a default value for boolean parameter, it can't display in the generated form, because the following code in **parameters-to-json-schema.js**:
`if (current.schema.default) schema.default = current.schema.default;`

2. If we pass a "false" to parameter for calling an API, the "false" can not be passed, due to the following in **oas-to-har.js**:
```
function formatter(values, param, type, onlyIfExists) {
  if (values[type][param.name]){
    return format.value(values[type][param.name]); 
  }
  ...
}

if (queryStrings && queryStrings.length) {
    queryStrings.forEach(queryString => {
      const value = formatter(formData, queryString, 'query', true);
      if (value) return;
    }
   ....
}
```

nadalfederer:
closes #149

### DIFF
--- a/packages/api-explorer/__tests__/lib/oas-to-har.test.js
+++ b/packages/api-explorer/__tests__/lib/oas-to-har.test.js
@@ -190,6 +190,26 @@ describe('path values', () => {
       ).log.entries[0].request.url,
     ).toBe('https://example.com/param-path/456');
   });
+
+  test('should add non-undefined values to the url', () => {
+    expect(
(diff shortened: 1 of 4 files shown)


domharrington:
Hey @nadalfederer, this looks great. Can you send me an email dom [at] readme.io so we can discuss your work with us further?

Issue #14, opened 2017-08-15
Support more parameter `style` values


domharrington:
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#style-values

nadalfederer:
@domharrington Is this suppose to be done here https://github.com/readmeio/api-explorer/blob/master/packages/api-explorer/src/lib/parameters-to-json-schema.js?
Can you give a little guide where it should be done and what to should be. I have read the docs, but I've not figured out where its located in code, or I didnt found the similar

domharrington:
I'd say it probably has to be done in here: https://github.com/readmeio/api-explorer/blob/master/packages/api-explorer/src/lib/oas-to-har.js. 

I think reading the docs is probably the best way to determine what this means: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#style-values

But in short, if someone defines an array for a JSON body parameter, then it's gunna be perfectly OK to send that through as a JSON array:

```json
{
  a: [1, 2, 3]
}
```

But if someone wants to have an array value as part of a header, or query, or path, these options are used to determine that. E.g. you may want it in the querystring as any one of the following:

```
/path?a=1,2,3
/path?a=1|2|3
/path?a=1%202%203
/path?a=1&a=2&a=3
```

These options are used to define that. TBH I don't think we've ever actually had anyone ask for this. 

This doc does explain in some more detail: https://swagger.io/docs/specification/serialization/

erunion:
Looked into this with @emilyskuo a couple months ago and it sent me into a rabbit hole of discovery with how Swagger UI does this. Turns out they use https://www.npmjs.com/package/swagger-client, which uses the following piece of code when building up parameters to be used in a HTTP module:

https://github.com/swagger-api/swagger-js/blob/master/src/execute/oas3/parameter-builders.js

I don't think we can outright use this piece of code because it's building up data to be appended to a `req` object, but we can potentially use the style serializer (https://github.com/swagger-api/swagger-js/blob/master/src/execute/oas3/style-serializer.js) that it uses. I'm not sure if adding this library as a dependency will add the full 2MB library to our Explorer bundle, so hopefully Webpack can treeshake it down to just this file that we load, but I'm not sure.

I still think that doing this work within `@readme/oas-to-har` is the way to go since that's what's handling all of this data for Try It request fetching and code snippets.

erunion:
we support this now!

PR #171, opened 2018-11-16
fix: missing 'properties' object in query param

### LINKED ISSUE #110: Query param type error
When using the “Object/Hash” query param type, after saving and viewing the docs I get the following error on the page:

image

nadalfederer:
closes #110
I've added an example to response-schema.json
Here how does it looked before fixing a bug
image
Here how does it look now
image
The bug occurred because we have passed incorrect json to `react-jsonschema-form`


domharrington:
I think this is a good start, but typing into the input, the object does not appear in the URL at all. Maybe we need to work on #14 as well in order to support more use cases.

erunion:
@nadalfederer Can you update this branch?

nadalfederer:
Sure, I'll do it soon

nadalfederer:
Sure, I'll do it soon

nadalfederer:
@erunion I've resolved conflicts

PR #176, opened 2018-11-30
Feed baseUrl to anchor

dok:
Enterprise styled URLs will have a baseUrl such as '/child/v1.0'
- Versions are optional
- Children are optional
- Language is optional

This PR will resolve and optionally add that URL structure to anchor tags when they render.

To test this, you must do two things
- Create custom body with `create-docs`. Uncomment the body variable.
- Change the `baseUrl` in demo to something else.

### DIFF
--- a/example/src/Demo.jsx
+++ b/example/src/Demo.jsx
@@ -31,6 +31,8 @@ function Demo({ fetchSwagger, status, docs, oas, oauth }) {
               'api-setting': Object.assign(extensions.defaults, oas),
             }}
             baseUrl={'/'}
+            // Uncomment this if you want to test enterprise-structured URLs
+            // baseUrl={'/child/v1.0/'}
             Logs={Logs}
             flags={{ correctnewlines: false }}
             // Uncomment this in for column layout
--- a/packages/api-explorer/src/block-types/Content.jsx
+++ b/packages/api-explorer/src/block-types/Content.jsx
@@ -12,9 +12,9 @@ const PropTypes = require('prop-types');
 
 const parseBlocks = require('../lib/parse-magic-blocks');
 
-const Loop = ({ content, column, flags }) => {
+const Loop = ({ content, column, flags, baseUrl }) => {
   const elements = content.map((block, key) => {
-    const props = { key, block, flags };
+    const props = { key, block, flags, baseUrl };
     switch (block.type) {
       case 'textarea':
         // eslint-disable-next-line react/no-array-index-key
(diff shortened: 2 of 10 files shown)


example/src/Demo.jsx
```diff
@@ -31,6 +31,8 @@ function Demo({ fetchSwagger, status, docs, oas, oauth }) {
               'api-setting': Object.assign(extensions.defaults, oas),
             }}
             baseUrl={'/'}
+            // Uncomment this if you want to test enterprise-structured URLs
+            // baseUrl={'/child/v1.0/'}
```
domharrington: If `baseUrl` ends with a slash, then the resulting URL has a double slash in the middle of it:

```
http://localhost:9966/child/v1.0//docs/link
```

I think a concatenation is good enough, as long as we can guarantee ReadMe will only pass URLs without the slash on the end.

example/src/Demo.jsx
```diff
@@ -31,6 +31,8 @@ function Demo({ fetchSwagger, status, docs, oas, oauth }) {
               'api-setting': Object.assign(extensions.defaults, oas),
             }}
             baseUrl={'/'}
+            // Uncomment this if you want to test enterprise-structured URLs
+            // baseUrl={'/child/v1.0'}
```
domharrington: If baseUrl has a slash on the end of it, then the resulting URL will have a double slash:

```
http://localhost:9966/child/v1.0//docs/link
```

I think this is okay, as long as we can guarantee that ReadMe will only ever pass in baseUrls with no slash on the end.

packages/api-explorer/src/block-types/Content.jsx
```diff
@@ -66,12 +66,12 @@ const Content = props => {
...
         <div className="hub-reference-left">
           <div className="content-body">
-            <Loop content={left} column="left" flags={props.flags} />
+            <Loop content={left} column="left" flags={props.flags} baseUrl={props.baseUrl} />
           </div>
         </div>
         <div className="hub-reference-right">
           <div className="content-body">
-            <Loop content={right} column="right" />
+            <Loop content={right} column="right" baseUrl={props.baseUrl} />
```
domharrington: I wonder why we don't pass in the flags here. That's probably a bug?

dok: I think so!

domharrington: Would you mind adding?

PR #178, opened 2018-12-06
Add sample responses to stripe style layout

mjcuva:
image

Added sample responses! Needed to add a class to fix text color since the sidebar is dark.

### DIFF
--- a/packages/api-explorer/src/ResponseSchema.jsx
+++ b/packages/api-explorer/src/ResponseSchema.jsx
@@ -1,5 +1,6 @@
 const React = require('react');
 const PropTypes = require('prop-types');
+const classNames = require('classnames');
 
 const Oas = require('./lib/Oas');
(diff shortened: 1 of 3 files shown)


packages/api-explorer/src/ResponseSchema.jsx
```diff
@@ -87,7 +88,12 @@ class ResponseSchema extends React.Component {
     if (!operation.responses || Object.keys(operation.responses).length === 0) return null;
     const schema = this.getSchema(operation);
     return (
-      <div className="hub-reference-response-definitions">
+      <div
+        className={classNames({
```
domharrington: Small nit, but you can do this to achieve the same:

```js
className={classNames('hub-reference-response-definitions', {
  dark: this.props.theme === 'dark',
})}
```

PR #179, opened 2018-12-12
Fix issue where default values wouldn't be saved into the formData until edited

domharrington:
You can test this locally here: http://localhost:9966/?selected=swagger-files%2Ftypes.json
Search for "default value" and see that it is already in the code sample.

This issue luckily got fixed upstream:
https://github.com/mozilla-services/react-jsonschema-form/pull/1034

So to bring that in I needed to update my fork of the module.

I would love to get off of my fork and get back onto the main release,
but there's some outstanding work on my PR which means we can't right now:
https://github.com/mozilla-services/react-jsonschema-form/pull/954

### DIFF
--- a/packages/api-explorer/__tests__/Params.test.jsx
+++ b/packages/api-explorer/__tests__/Params.test.jsx
@@ -234,7 +234,7 @@ describe('readOnly', () => {
   });
 });
 
-test('defaults should be applied on first render', (done) => {
+test('defaults should be applied on first render', done => {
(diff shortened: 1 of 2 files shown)


domharrington:
We should deprecate the x-send-defaults option if it's no longer necessary: https://readme.readme.io/docs/swagger-extensions#section-send-defaults

mjcuva:
How was this fixed? Should there be a change in package.json as well?

PR #185, opened 2018-12-28
Selected example

dok:
https://app.asana.com/0/681252538274980/801649456511557/f

### Currently a work in progress

Problem: Selected manual code samples does not change for everything else, once code sample is selected.
- There is also the issue of duplicate languages; if a language is specified twice, what will determine what is selected?

Solution: Use top-level state for the CodeSample component, to manage selection.
- Using the prop `.language` is not enough to determine what is selected. There may be a CodeSample with examples `TypeScript` and `JavaScript` (names) that have language of `JavaScript` specified. Because of this, we need to be more specific.

How to debug: 
- Uncomment [these](https://github.com/readmeio/api-explorer/blob/master/packages/api-explorer/lib/create-docs.js#L51-L62) lines

Question
- How should we persist this data? Should we save it in cookies, like we do with the prop `language`? @domharrington

### DIFF
--- a/packages/api-explorer/package-lock.json
+++ b/packages/api-explorer/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@readme/api-explorer",
-  "version": "2.6.0",
+  "version": "3.0.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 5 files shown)


domharrington:
I can see how the name/language problem could come about. It's slightly annoying that this is only for manual API examples.

I like the current implementation because it's just a single language that's selected. Can we not just attempt to select the first language of that type, and if that's not possible then just select the first example? 

I'm aware that this may not be 100% correct, and if they have one endpoint with 2 random languages and they expect the 2nd one to be selected on refresh, then it's not gunna work. Maybe this is enough though?

The alternative is storing some kind of map somewhere like this:

```
{
  <_id of endpoint>: { language: 'javascript', name: 'typescript' }
}
```

Which will get very unwieldy very quickly.

dok:
I understand your frustration, as it annoyed me quite a bit as well to know that `language` is not a unique property selecting a code block for a `language`.

But I don't think selecting the language of the type is an acceptable solution. We've allowed users to create code blocks with duplicate languages that are identifiable by the property `name`. The current implementation is simple, but unfortunately, that's not gonna satisfy customer's requirements.

Actually, the previous implementation of state had an internal state within the CodeBlock component that did store the map/object of `language` and `name`.

PR #186, opened 2019-01-18
Feature/typescript

dok:
Part 2 of https://app.asana.com/0/699361672159104/801649456511557

- Add typescript to modes
- Export modes so that ReadMe can digest it
- Select 0th item if `selectedExamples` is not found

### DIFF
--- a/packages/api-explorer/src/CodeSample.jsx
+++ b/packages/api-explorer/src/CodeSample.jsx
@@ -25,7 +25,10 @@ class CodeSample extends React.Component {
 
   getKey(example, index) {
     const key = `${example.language}-${index}`;
-    const selected = this.state.selectedExample === example;
+    let selected = this.state.selectedExample === example;
+    if (!this.state.selectedExample && index === 0) {
+      selected = true;
+    }
     return { key, selected };
   }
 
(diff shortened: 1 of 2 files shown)


packages/syntax-highlighter/index.js
```diff
@@ -11,3 +11,5 @@ module.exports = (code, lang, opts = { dark: false, tokenizeVariables: false })
   );
 
 module.exports.uppercase = require('./uppercase');
+
+module.exports.modes = modes;
```
domharrington: Have you tried importing this into readme yet? I feel like you may have problems because our regular webpack config doesnt have .jsx support. I would maybe add it as a separate modes.js file, then we can require it as `require('@readme/syntax-highlighter/modes')` without needing to touch jsx files.

packages/api-explorer/src/CodeSample.jsx
```diff
@@ -25,7 +25,10 @@ class CodeSample extends React.Component {
 
   getKey(example, index) {
     const key = `${example.language}-${index}`;
-    const selected = this.state.selectedExample === example;
+    let selected = this.state.selectedExample === example;
```
domharrington: Can you add some tests for this? I'm struggling to parse the logic behind this in my head, so having a test would be great!

dok:
@domharrington I've added a test for default selection and moved the modes configs to a seperate file

PR #196, opened 2019-04-02
Remove trailing slashes from paths and header parameter improvements

kanadgupta:
Adds the following (with corresponding tests):
- Removes trailing slashes from the paths when operations are created
- Overrides Accept header types found in response definitions if an Accept header is sent in the form data
- Added a `$ref` lookup to path operation parameters

Asana tasks:
- https://app.asana.com/0/681252538274980/1108254356757915/f
- https://app.asana.com/0/681252538274980/977151049856241/f

### DIFF
--- a/packages/api-explorer/src/lib/Oas.js
+++ b/packages/api-explorer/src/lib/Oas.js
@@ -5,7 +5,8 @@ class Operation {
   constructor(oas, path, method, operation) {
     Object.assign(this, operation);
     this.oas = oas;
-    this.path = path;
+    // strips trailing slash
+    this.path = path.endsWith('/') ? path.slice(0, -1) : path;
     this.method = method;
   }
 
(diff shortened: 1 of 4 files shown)


packages/api-explorer/src/lib/Oas.js
```diff
@@ -5,7 +5,8 @@ class Operation {
   constructor(oas, path, method, operation) {
     Object.assign(this, operation);
     this.oas = oas;
-    this.path = path;
+    // strips trailing slash
+    this.path = path.endsWith('/') ? path.slice(0, -1) : path;
```
domharrington: Yo! `String.prototype.endsWith` isn't supported in ie11. Can you check to see if this works in ie11? I can give you access to our browserstack account. I'm hoping that we've got @babel/env setup correctly here: https://github.com/readmeio/api-explorer/blob/7963c659cd927486abf0d15b4f7a4f8622722aca/.babelrc#L3-L7 but it's possible that we need to add an exclusion to allow ie11.

kanadgupta: I'd be happy to test it in IE11 or just refactor this to something more universally compatible too. What would you prefer?

PR #202, opened 2019-04-04
strip out the string format that we do not support.

### LINKED ISSUE #197: Do not throw an error on invalid string format
If someone provides the following inside of a request body:

```
{
  type: 'string',
  format: 'some-unknown-format'
}
```

If it is not one of our known formats: https://github.com/readmeio/api-explorer/blob/7963c659cd927486abf0d15b4f7a4f8622722aca/packages/api-explorer/src/Params.jsx#L50-L61 then our form renderer returns an unrecoverable error. We should strip out the format if it is one that we do not support.

I suspect that we may be able to do this somewhere in here: https://github.com/readmeio/api-explorer/blob/7963c659cd927486abf0d15b4f7a4f8622722aca/packages/api-explorer/src/form-components/SchemaField.jsx

nadalfederer:
closes #197
`before`
image
`after`
image

### DIFF
--- a/example/swagger-files/petstore.json
+++ b/example/swagger-files/petstore.json
@@ -222,8 +222,7 @@
           "in": "formData",
           "description": "Updated name of the pet",
           "required": false,
-          "type": "string",
-          "format": "some-unknown-format"
+          "type": "string"
         }, {
           "name": "status",
           "in": "formData",
--- a/example/swagger-files/petstore.json
+++ b/example/swagger-files/petstore.json
@@ -222,7 +222,8 @@
           "in": "formData",
           "description": "Updated name of the pet",
           "required": false,
-          "type": "string"
+          "type": "string",
+          "format": "some-unknown-format"
         }, {
           "name": "status",
           "in": "formData",
--- a/packages/api-explorer/src/form-components/SchemaField.jsx
+++ b/packages/api-explorer/src/form-components/SchemaField.jsx
@@ -9,21 +9,8 @@ function getDefaultNumFormat(type) {
   return '';
 }
 
-function doesFormatExist(type, format) {
-  const availableFormats = [
-    'int64',
-    'int32',
-    'double',
-    'float',
-    'binary',
-    'byte',
-    'string',
-    'uuid',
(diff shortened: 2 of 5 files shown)


packages/api-explorer/src/form-components/SchemaField.jsx
```diff
@@ -28,6 +47,8 @@ function getCustomType(schema) {
 }
 
 function SchemaField(props) {
+  if (!doesFormatExist(props.schema.type, props.schema.format)) props.schema.format = undefined;
```
domharrington: Can you add a test for this case? Tests are written via this file: https://github.com/readmeio/api-explorer/blob/master/packages/api-explorer/__tests__/Params.test.jsx#L50-L63

packages/api-explorer/src/form-components/SchemaField.jsx
```diff
@@ -9,6 +9,25 @@ function getDefaultNumFormat(type) {
   return '';
 }
 
+function doesFormatExist(type, format) {
+  const availableFormats = [
```
domharrington: You can dynamically get a list of these from `props.registry.widgets` inside of this function: https://github.com/readmeio/api-explorer/pull/202/files#diff-ca60764ec4fc812c3c70557738a148f0R49

This is the result of a merge of the widgets that we pass in + the defaults. This will save us from having to update this list in future if we add more. Alternatively, we could export this object from here https://github.com/readmeio/api-explorer/blob/master/packages/api-explorer/src/Params.jsx#L50-L61 and utilize it in here.

example/swagger-files/petstore.json
```diff
@@ -222,7 +222,8 @@
           "in": "formData",
           "description": "Updated name of the pet",
           "required": false,
-          "type": "string"
+          "type": "string",
```
domharrington: Can you instead add this here: https://github.com/readmeio/api-explorer/blob/master/example/swagger-files/types.json#L57-L60?

nadalfederer:
thanks for the review, I fixed current bugs

PR #203, opened 2019-04-04
Add support for string format: url

### LINKED ISSUE #201: Add support for string `format: "url"`
There's a URL widget in the form module we use: https://github.com/mozilla-services/react-jsonschema-form/blob/889871f90c6b03aed4df431d92fa60de10be24a6/src/components/widgets/URLWidget.js

We should expose this to our end users here: https://github.com/readmeio/api-explorer/blob/7963c659cd927486abf0d15b4f7a4f8622722aca/packages/api-explorer/src/Params.jsx#L50-L63

This results in an `<input type='url' />`

nadalfederer:
closes #201

### DIFF
--- a/example/swagger-files/petstore.json
+++ b/example/swagger-files/petstore.json
@@ -229,6 +229,13 @@
           "description": "Updated status of the pet",
           "required": false,
           "type": "string"
+        }, {
+          "name": "url image",
+          "in": "formData",
+          "description": "pet's image url",
+          "required": false,
+          "type": "string",
+          "format": "url"
         }],
         "responses": {
           "405": {
--- a/example/swagger-files/petstore.json
+++ b/example/swagger-files/petstore.json
@@ -229,13 +229,6 @@
           "description": "Updated status of the pet",
           "required": false,
           "type": "string"
-        }, {
-          "name": "url image",
-          "in": "formData",
-          "description": "pet's image url",
-          "required": false,
-          "type": "string",
-          "format": "url"
         }],
         "responses": {
           "405": {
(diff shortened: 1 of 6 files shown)


nadalfederer:
I dont know what is wrong with `codeclimate`. `npm t` does not showing any error.

packages/api-explorer/src/form-components/URLWidget.jsx
```diff
@@ -0,0 +1,14 @@
+const extensions = require('@readme/oas-extensions');
+
+const URLWidget = require('react-jsonschema-form/lib/components/widgets/URLWidget').default;
+
+function createURLWidget(oas) {
+  const explorerEnabled = oas[extensions.EXPLORER_ENABLED];
+
+  // Return a function that renders null when the explorer is disabled
+  if (!explorerEnabled) return () => null;
```
domharrington: Can you add a test for this case like this one: https://github.com/readmeio/api-explorer/blob/master/packages/api-explorer/__tests__/Params.test.jsx#L212-L222

You may have to modify the fixture file to add a string with `format: url` to test this.

example/swagger-files/petstore.json
```diff
@@ -229,6 +229,13 @@
           "description": "Updated status of the pet",
           "required": false,
           "type": "string"
+        }, {
```
domharrington: Instead of adding this here, can you add it to this file like these ones: https://github.com/readmeio/api-explorer/blob/master/example/swagger-files/types.json#L40-L60

nadalfederer:
thanks for review, I've fixed current bugs

Issue #104, opened 2018-02-15
requestBody with oneOf schemas produces an error


ahannigan:
When using a valid OpenAPI 3.0.0 specification with a requestBody that contains a oneOf schema like below:

```yaml
requestBody:
  content:
    application/json:
      schema:
        oneOf:
          - $ref: '#/components/schemas/One'
          - $ref: '#/components/schemas/Two'
          - $ref: '#/components/schemas/Three'
```

the following error is produced:

`Unsupported field schema for field root: Unknown field type undefined.`

followed by the oneOf JSON blob, rather than rendering the available request bodies.

uppal101:
Unfortunately the library in which we are using does not have the support for this yet once it does we will support it. https://github.com/mozilla-services/react-jsonschema-form/pull/417

spifd:
Any update supporting this? Apparently, react-jsonschema-form still doesn't support it. This is a strong limitation when handling advanced JSON structure in API.

epicfaace:
We've recently merged in a PR that will support it in react-jsonschema-form, so expect it in the next release: mozilla-services/react-jsonschema-form#1133

ekeren:
Hey, any idea when is the next release?

epicfaace:
The above PR has already been released in [1.2.1](https://github.com/mozilla-services/react-jsonschema-form/releases/tag/v1.2.1), @ekeren

domharrington:
This is now fixed: https://preview.readme.io/?selected=swagger-files%2Fone-of.json

PR #204, opened 2019-04-05
add empty `properties` key to type 'object'

### LINKED ISSUE #174: Explorer doesn't render when the object reference is empty
Look at the following code: 

```
/endpoints/{deploymentName}/datavec/{transformName}/{versionName}/transformprocess:
    post:
      operationId: transformprocess_post
      summary: Sets the deployed (CSV or Image) transform process through the provided JSON string
      tags:
      - Transform
      consumes:
      - application/json
      produces:
      - application/json
      parameters:
      - name: deploymentName
        in: path
        description: Name of the deployment group
        required: true
        type: string
      - name: versionName
        in: path
        description: Version name of the endpoint. The default value is "default"
        required: true
        default: default
        type: string
      - name: transformName
        in: path
        description: ID or name of the deployed transform
        required: true
        type: string
      - in: body
        name: TransformProcess
        description: The transform process to set
        schema:
          type: object
      responses:
        200:
          description: transform process set successfully
          schema:
            type: object
        500:
          description: internal server error
```

It gives the following output in the api-explorer:

capture

By saying: `Invalid root object field configuration:Cannot convert undefined or null to object.`

**Proposal**
Maybe a better way to deal with this case is to show nothing under the `BODY PARAMS` header.

nadalfederer:
closes #174

### DIFF
--- a/example/swagger-files/types.json
+++ b/example/swagger-files/types.json
@@ -131,6 +131,9 @@
                   "object with `additionalProperties: true`": {
                     "type": "object",
                     "additionalProperties": true
+                  },
+                  "object without `properties`": {
+                    "type": "object"
                   }
                 }
               }
--- a/packages/api-explorer/__tests__/Params.test.jsx
+++ b/packages/api-explorer/__tests__/Params.test.jsx
@@ -243,3 +243,15 @@ test('defaults should be applied on first render', done => {
 
   renderParams({ type: 'string', default: defaultValue }, { onChange });
 });
+
+describe('should not contains error message when property key missing in object type', () => {
+  test('should make `readOnly` properties hidden', () => {
+    expect(
+      mount(
+        <div>
+          <Params {...props} operation={oas.operation('/pet/{petId}', 'post')} />
+        </div>,
+      ).html().includes('Invalid empty object object field configuration'),
+    ).toBe(false);
+  });
+})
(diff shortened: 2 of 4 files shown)


nadalfederer:
@domharrington Can you check this. Is it the right approach to do this task?

domharrington:
Hey! So it looks like there's been a fix for this in the upstream form module which defaults it to `{}` if it's not set. This would be a much simpler solution if we can just bring this in from upstream: https://github.com/mozilla-services/react-jsonschema-form/blob/f9d4c63cac24146e522528d33eb04f465464e052/src/components/fields/ObjectField.js#L201

We're currently using a fork of this module: https://github.com/readmeio/api-explorer/blob/cf83d74fb7d18063f21906e232db24804acf8095/packages/api-explorer/package.json#L22 but I have to merge in the latest master anyway to get the `anyOf/oneOf` support for this: https://github.com/readmeio/api-explorer/issues/104

So maybe it will be fixed by default when we pull that in?

nadalfederer:
Yeah, the latest version of `react-jsonschema-form` fixed that bug

domharrington:
I just bumped this in here: https://github.com/readmeio/api-explorer/tree/fix/do-not-show-json-on-empty-object-declaration

Can you merge this branch into yours and repush to update this PR?

nadalfederer:
@domharrington I've updated this branch

PR #205, opened 2019-04-05
Fix/height

dok:
Adds height to root container when height is too short, https://preview.readme.io/?selected=swagger-files%2Fboolean.json

There is the fix in my local machine:

Screen Shot 2019-04-30 at 12 54 57 PM
http://localhost:9966/?selected=swagger-files%2Fboolean.json

Fix utilizes an npm package called `react-resize-detector` to check for resize events

Screen Shot 2019-04-30 at 6 49 05 PM


dok:
@kanadgupta isn't `dist/index.js` supposed to change as `src/*` changes?

kanadgupta:
> @kanadgupta isn't `dist/index.js` supposed to change as `src/*` changes?

Yes, but we do a separate update of `dist/index.js` when building for deployment to prod and npm, which is why we don't change it in our PRs.

erunion:
Closing this out in favor of a pure CSS fix in https://github.com/readmeio/api-explorer/pull/207

PR #206, opened 2019-04-06
Automatically default `type` to object if no type set

### LINKED ISSUE #200: Automatically default `type` to object if no type set
This is kinda the other side of this issue: #199.

Some people pass `properties` but do not tell us that it's an object. We should default `type` to `object` if they've not provided one.

This is what happens if you remove `type: object` from a JSON schema:

image

nadalfederer:
closes #200

### DIFF
--- a/packages/api-explorer/__tests__/Params.test.jsx
+++ b/packages/api-explorer/__tests__/Params.test.jsx
@@ -243,3 +243,15 @@ test('defaults should be applied on first render', done => {
 
   renderParams({ type: 'string', default: defaultValue }, { onChange });
 });
+
+test('should set type object if it missing', () => {
+  expect(
+    mount(
+      <div>
+        <Params {...props} operation={oas.operation('/pet/{petId}', 'post')} />
+      </div>,
+    )
+      .html()
+      .includes('Unsupported field'),
+  ).toBe(false);
+});
--- a/packages/api-explorer/__tests__/fixtures/petstore/oas.json
+++ b/packages/api-explorer/__tests__/fixtures/petstore/oas.json
@@ -289,6 +289,19 @@
                   "status": {
                     "description": "Updated status of the pet",
                     "type": "string"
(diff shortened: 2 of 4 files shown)


nadalfederer:
@domharrington this ready. Can you check it when you available

domharrington:
I don't particularly like this solution, I don't think it would work for $refs which makes me think it has to be done at the form layer. I've added an issue upstream for this to see if it's something we can fix there: https://github.com/mozilla-services/react-jsonschema-form/issues/1265

I had some limited success locally making a fix here:
https://github.com/readmeio/api-explorer/blob/b32a2146737c11813bd1b222a137de61854414b3/packages/api-explorer/src/Params.jsx#L73-L77

```
fields={{
  DescriptionField,
  ArrayField,
  SchemaField,
  // This is a kinda hacky way for us
  // to default when `type` is undefined
  // to just attempt to render an object
  undefined: UndefinedField,
}}
```

With UndefinedField containing this:

```

const ObjectField = require('react-jsonschema-form/lib/components/fields/ObjectField')
  .default;

function UndefinedField(props) {
  return (
    <ObjectField
      {...props}
      schema={Object.assign({}, props.schema, { type: 'object' })}
    />
  );
}

module.exports = UndefinedField;
```

But it resulted in too many labels showing up. It may be the best we can do for now.

erunion:
This is no longer a bug, but there's still some work that could be done here.

Currently, an object without a `type` looks like this:

Screen Shot 2019-08-12 at 10 31 02 AM

So we're still showing too many labels, but I think that this is a RJSF bug that should be resolved there instead because with a type that object renders just fine:

Screen Shot 2019-08-12 at 10 32 27 AM

erunion:
Since having a `type` no longer crashes rendering, I'm filing a new bug for the too-many-labels issue instead. https://github.com/readmeio/api-explorer/issues/274

PR #208, opened 2019-04-14
feat: support multiple response types

### LINKED ISSUE #198: Support multiple response types
We currently display response schemas on the right hand side:

image

https://preview.readme.io/?selected=swagger-files%2Fexamples.json

This comes from this swagger file: https://github.com/readmeio/api-explorer/blob/master/example/swagger-files/examples.json#L21-L30

I misread the spec and assumed that examples were always at `examples.response.value` BUT it appears as though you can do `examples.${any-string}.value` to define multiple examples

Here’s the relevant part of the spec: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#media-type-object

And here’s an example of it being used as intended with multiple example responses: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#media-type-examples

Right now we only support one example response per status code. So i think the proper solution would be to add some form of dropdown underneath the status code bar to allow you to switch between example responses.

Some additional requirements:
- If there's only one response, don't show a dropdown
- Automatically select/show the first response type

nadalfederer:
closes #198
@domharrington this is ready. Can you check it when you available
https://github.com/readmeio/api-explorer/pull/208/files#diff-e7f8868abb312f633724097c3a090b40R61
I fixed a broken/wrong test here which came from the master


nadalfederer:
Peek 2019-04-14 22-01

packages/api-explorer/__tests__/Params.test.jsx
```diff
@@ -58,8 +58,8 @@ test('boolean should render as <select>', () => {
   const select = params.find('.field-boolean select');
 
   expect(select.length).toBe(1);
-  expect(select.find('option').length).toBe(3);
-  expect(select.find('option').map(el => el.text())).toEqual(['', 'true', 'false']);
+  expect(select.find('option').length).toBe(2);
```
domharrington: This has been fixed in master now! It was due to an upstream change in the form module. Re-merging master will fix this test. Here's the commit: https://github.com/readmeio/api-explorer/commit/9d2e001346774a6792861bebad3b2618f123d6fc

nadalfederer: I've reverted this test and merged master into this branch.

packages/api-explorer/src/Example.jsx
```diff
@@ -14,7 +14,68 @@ const Oas = require('./lib/Oas');
...
+}
+
+function showExamples(examples, setResponseType, responseType) {
+  let responseTypeCopy = responseType;
+  if (!responseTypeCopy && examples[0]) responseTypeCopy = examples[0].label;
+
+  return (
+    <div>
+      <select
+        style={{ backgroundColor: '#3d434a' }}
```
erunion: It would probably be best if this were a class name instead. Can you add a class for this into `example/bundle-hub2.css`? I'll make sure we do the same to our production CSS.

nadalfederer: I've added this

PR #209, opened 2019-04-19
Improve handling of baseUrl

kanadgupta:
Create baseUrl [context](https://reactjs.org/docs/context.html) and revert prop passing in #176.

`baseUrl` isn't properly being passed down for certain fields such as parameter descriptions, excerpts, and tables. Taking advantage of contexts should eliminate the need for all the props passing that was being done.

### DIFF
--- a/packages/api-explorer/src/index.jsx
+++ b/packages/api-explorer/src/index.jsx
@@ -5,6 +5,7 @@ const extensions = require('@readme/oas-extensions');
 const VariablesContext = require('@readme/variable/contexts/Variables');
 const OauthContext = require('@readme/variable/contexts/Oauth');
 const GlossaryTermsContext = require('@readme/markdown/contexts/GlossaryTerms');
+const BaseUrlContext = require('@readme/markdown/contexts/BaseUrl');
 const SelectedAppContext = require('@readme/variable/contexts/SelectedApp');
 
 const ErrorBoundary = require('./ErrorBoundary');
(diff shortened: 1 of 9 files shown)


packages/api-explorer/src/index.jsx
```diff
@@ -82,25 +83,27 @@ class ApiExplorer extends React.Component {
...
+                    <SelectedAppContext.Provider value={this.state.selectedApp}>
+                      <Doc
+                        key={doc._id}
+                        doc={doc}
+                        oas={this.getOas(doc)}
+                        setLanguage={this.setLanguage}
+                        flags={this.props.flags}
+                        user={this.props.variables.user}
+                        Logs={this.props.Logs}
+                        baseUrl={this.props.baseUrl.replace(/\/$/, '')}
```
domharrington: Does this still need to be here?

kanadgupta: Yes I believe so - we need to pass down the baseUrl prop into two files for [docs](https://github.com/readmeio/api-explorer/blob/fix/baseUrl-in-anchors/packages/api-explorer/src/Doc.jsx) and [logs](https://github.com/readmeio/api-explorer/blob/fix/baseUrl-in-anchors/packages/api-logs/index.jsx).

packages/markdown/__tests__/__snapshots__/index.test.js.snap
```diff
@@ -11,10 +11,10 @@ exports[`anchors 1`] = `
 `;
 
 exports[`anchors with baseUrl 1`] = `
-"<p><a href=\\"/child/v1.0/docs/slug\\" target=\\"_self\\" class=\\"doc-link\\" data-sidebar=\\"slug\\">doc</a><br/>
-<a href=\\"/child/v1.0/reference-link/slug\\" target=\\"_self\\">ref</a><br/>
-<a href=\\"/child/v1.0/blog/slug\\" target=\\"_self\\">blog</a><br/>
-<a href=\\"/child/v1.0/page/slug\\" target=\\"_self\\">page</a><br/>
+"<p><a href=\\"/child/v1.0/docs/slug\\" target=\\"_self\\" class=\\"doc-link\\" data-sidebar=\\"slug\\">doc</a><br>
```
domharrington: Weird that enzyme does this but OK! 🤨

PR #211, opened 2019-04-27
Add support for common parameters

### LINKED ISSUE #210: Add support for common parameters
We currently do not support common parameters. (https://swagger.io/docs/specification/describing-parameters/#common-parameters)

Workaround is to add them inline for each endpoint for now

Taken from canny: https://readmeio.canny.io/feature-requests/p/add-support-for-common-parameters

----------

Normally you add parameters in Open API like so:

```js
{
  paths: {
    '/path': {
      get: {
        parameters: []
      }
    }
  }
}
```

But it is supported for you to define them like this:

```js
{
  paths: {
    '/path': {
      parameters: []
      get: {
      }
    }
  }
}
```

We should add support for these.

nadalfederer:
closes #210

### DIFF
--- a/packages/api-explorer/__tests__/lib/parameters-to-json-schema.test.js
+++ b/packages/api-explorer/__tests__/lib/parameters-to-json-schema.test.js
@@ -474,7 +474,7 @@ test('it should add common parameter to path params', () => {
   ).toEqual(oas.paths['/pet/{petId}'].parameters[0].description);
 });
 
-test('it should overrides path-level parameters on the operation level', () => {
+test('it should override path-level parameters on the operation level', () => {
   const oas = {
     paths: {
       '/pet/{petId}': {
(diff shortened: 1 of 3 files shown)


nadalfederer:
@domharrington Hi, 
Screenshot from 2019-04-27 13-53-56
There is an addition `method` that appears because of new 'parameters' under the path. It is supposed to be hidden, right?
image

nadalfederer:
I've fixed this :arrow_up: 
Paths may have an optional short `summary` and  `description`, I've excluded by filtering by methods which OpenAPI 3.0 supports  - get, post, put, patch, delete, head, options, and trace.
Its describes here https://swagger.io/docs/specification/paths-and-operations/

nadalfederer:
Thanks for the review. I've resolved conflicts.

PR #212, opened 2019-04-28
Fix style issue when "array of objects" contains another object

### LINKED ISSUE #193: Fix style issue when "array of objects" contains another object
Replace this object here: https://github.com/readmeio/api-explorer/blob/master/example/swagger-files/types.json#L180

With this:

```json
"array of objects": {
  "type": "array",
  "items": {
    "type": "object",
    "properties": {
      "id": {
        "type": "integer",
        "format": "int64"
      },
      "name": {
        "type": "string"
      },
      "object": {
        "type": "object",
        "properties": {
          "a": {
            "type": "string"
          }
        }
      }
    }
  }
}
```

This incorrectly outputs an "array of objects" label around the inner object:

image

It should just display "object" in the :after element.

nadalfederer:
closes #193
image
This is ready

### DIFF
--- a/packages/api-explorer/api-explorer.css
+++ b/packages/api-explorer/api-explorer.css
@@ -313,7 +313,7 @@ form.rjsf button[disabled] {
 
 .field.field-array-of-object > legend::after {
   color: rgba(0, 0, 0, 0.6);
-  content: 'array of objects' !important;
+  content: 'array of objects';
   background: #f2f2f2;
   border-radius: 3px;
   font-family: 'Proxima Nova', sans-serif;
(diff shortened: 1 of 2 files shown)


packages/api-explorer/api-explorer.css
```diff
@@ -323,6 +323,18 @@ form.rjsf button[disabled] {
   padding: 3px 6px;
 }
 
+fieldset[id^='root_array of objects_'] > legend::after {
```
rafegoldberg: looks good to me! (tho this change may be superseded by my UI updates in #228)

PR #214, opened 2019-05-13
Do not strip trailing slashes and instead create unique key for path URL

kanadgupta:
In #196, we stripped trailing slashes from path URLs, primarily because it would throw a React key uniqueness error.

Screenshot 2019-03-29 12 27 03

Instead of simply using the value of each part to designate the key, this PR will instead create a separate key for each part of the path URL that appends a value count. For example, if there are two values of `/`, they will have respective key values of `/-0` and `/-1`.

### DIFF
--- a/packages/api-explorer/src/PathUrl.jsx
+++ b/packages/api-explorer/src/PathUrl.jsx
@@ -9,11 +9,20 @@ const { Operation } = Oas;
 const extensions = require('@readme/oas-extensions');
 
 function splitPath(path) {
+  const keys = [];
   return path
     .split(/({.+?})/)
     .filter(Boolean)
     .map(part => {
       return { type: part.match(/[{}]/) ? 'variable' : 'text', value: part.replace(/[{}]/g, '') };
+    })
+    .map(part => {
+      // To ensure unique keys, we're going to create a key with
+      // the value concatenated to the count of the value.
+      const keyCount = keys.filter(key => key === part.value).length;
+      keys.push(part.value);
+      part.key = `${part.value}-${keyCount}`;
+      return part;
     });
 }
 function PathUrl({
@@ -73,7 +82,7 @@ function PathUrl({
             <span className="definition-url">
               <span className="url">{oas.url()}</span>
               {splitPath(operation.path).map(part => (
-                <span key={part.value} className={`api-${part.type}`}>
+                <span key={part.key} className={`api-${part.type}`}>
                   {part.value}
                 </span>
               ))}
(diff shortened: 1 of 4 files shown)


packages/api-explorer/src/PathUrl.jsx
```diff
@@ -9,11 +9,20 @@ const { Operation } = Oas;
 const extensions = require('@readme/oas-extensions');
 
 function splitPath(path) {
+  const keys = [];
```
domharrington: You could use a Set for this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

Looks like we can use it in ie11: https://kangax.github.io/compat-table/es6/#test-Set

packages/api-explorer/src/PathUrl.jsx
```diff
@@ -9,11 +9,20 @@ const { Operation } = Oas;
...
     .filter(Boolean)
     .map(part => {
       return { type: part.match(/[{}]/) ? 'variable' : 'text', value: part.replace(/[{}]/g, '') };
+    })
+    .map(part => {
+      // To ensure unique keys, we're going to create a key with
+      // the value concatenated to the count of the value.
+      const keyCount = keys.filter(key => key === part.value).length;
+      keys.push(part.value);
+      part.key = `${part.value}-${keyCount}`;
```
domharrington: You could also, instead of finding a count, use the `i` value of the current iteration and add that.

```
.map((part, i) => {
```

That may negate the need for the array/Set altogether?

packages/api-explorer/src/lib/Oas.js
```diff
@@ -5,8 +5,7 @@ class Operation {
   constructor(oas, path, method, operation) {
     Object.assign(this, operation);
     this.oas = oas;
-    // strips trailing slash
-    this.path = path.endsWith('/') ? path.slice(0, -1) : path;
```
domharrington: Surprised there was no test for this in Oas.test.js?

packages/api-explorer/src/PathUrl.jsx
domharrington: Ignore this. I misunderstood.

packages/api-explorer/src/lib/Oas.js
kanadgupta: There was! I removed it here: https://github.com/readmeio/api-explorer/pull/214/files#diff-118b2290dc906ad945edd96c4401c538L16

packages/api-explorer/src/PathUrl.jsx
kanadgupta: Huh hadn't thought of that! I guess there's no reason to use a unique index specific to the value itself. I'll simplify that!

PR #220, opened 2019-05-31
App selections should update current auth states

domharrington:
* [x] Changing a selection in the Logs dropdown should update current auth states and code sample.
* [x] A dropdown of the user's apps/API keys should be put into the auth box and experience the same state changes as the Logs dropdown.
* [x] Password inputs on Basic HTTP auth are now `input[type=password]` inputs.

g6JHsGGj7c

#### 🚧 Revised design
Screen Shot 2019-11-25 at 10 42 22 AM


mjcuva:
Two issues with this for the next time @domharrington and I talk about this (forgetting we've done this twice now):
* API Key isn't editable in the auth dropdown
  * We should make it editable if logged out/no api keys in jwt
  * We should make it a picker for the api key if logged in with jwt
* Logs dropdown doesn't update api key (or variables)

packages/api-logs/index.jsx
```diff
@@ -88,7 +88,7 @@ class Logs extends React.Component {
   }
 
   onSelect(event) {
-    this.changeGroup(event.target.value);
+    this.changeGroup(event.target.value, event.target.options[event.target.selectedIndex].text);
```
erunion: Since `getAuth` asks for a `selectedApp` param, it makes more sense to push back the group name instead of the group ID/api key to the explorer.

packages/api-explorer/src/Doc.jsx
```diff
@@ -353,7 +356,6 @@ Doc.propTypes = {
   }),
   auth: PropTypes.shape({}).isRequired,
   baseUrl: PropTypes.string,
-  changeGroup: PropTypes.func.isRequired,
```
erunion: Renamed this throughout so it's clearer that this is intended for an `onChange` event.

packages/api-logs/index.jsx
```diff
@@ -88,7 +88,7 @@ class Logs extends React.Component {
   }
 
   onSelect(event) {
-    this.changeGroup(event.target.value);
+    this.onGroupChange(event.target.value, event.target.options[event.target.selectedIndex].text);
```
erunion: Since `getAuth` asks for a `selectedApp` param, it makes more sense to push back the group name as well instead of only the group ID/api key to the explorer.

erunion:
@domharrington @mjcuva I'm working on adding some more unit tests, but this is ready for review.

@rafegoldberg We might want to change the font color of the dropdown in the `AuthBox` to match the color of it in the `Logs` module.

erunion:
With multiple securities on an operation, it kind of looks like garbage. @madisonyocum @mjcuva @rafegoldberg ideas?

Screen Shot 2019-11-22 at 9 59 06 AM

Also changing those dropdowns right now in that example updates both securities. That seems wrong.

packages/api-explorer/src/AuthBox.jsx
```diff
@@ -5,14 +5,40 @@ const { Operation } = require('oas');
...
+  authInputRef,
+  group,
+  groups,
+  oauth,
+  onChange,
+  onGroupChange,
+  onSubmit,
+  operation,
+}) {
+  function onSelect(event) {
```
domharrington: This will make this code a little cleaner:

```suggestion
  function onSelect({ target }) {
```

packages/api-explorer/src/lib/get-auth.js
```diff
@@ -32,24 +32,30 @@ function getSingle(user, scheme = {}, selectedApp = false) {
...
-            [scheme]: getSingle(user, {
-              ...oas.components.securitySchemes[scheme],
-              _key: scheme,
-            }),
+            [scheme]: getSingle(
+              user,
+              {
+                ...oas.components.securitySchemes[scheme],
+                _key: scheme,
+                selectedApp,
```
domharrington: Is this needed?

packages/api-logs/index.jsx
domharrington: Can the group name be inferred from the value higher up the stack? Or does it make sense to always have to pass in the pair here?

packages/api-logs/package.json
```diff
@@ -15,7 +15,7 @@
     "pretest": "npm run lint && npm run inspect && npm run prettier",
     "prettier": "prettier --list-different --write \"./**/**.{js,jsx}\"",
     "test": "jest --coverage --runInBand",
-    "watch": "webpack -w"
+    "watch": "webpack -w --progress"
```
domharrington: 👀

erunion:
Talked offline with @mjcuva about the dupe dropdown issue and we're going to play with moving the dropdown into the top right of the authbox and only show it once.

Screen Shot 2019-11-22 at 3 09 12 PM

packages/api-explorer/src/lib/get-auth.js
erunion: Yeah, `getSingle` needs to know the current app so it can pick out the right key.

erunion:
@domharrington @rafegoldberg @mjcuva @madisonyocum This is good for review again. The design has slightly changed per our conversation last week to move the groups dropdown up into the first auth header:

69568681-eae74400-0f70-11ea-8b0f-924d4bc1670a

Functionality remains the same though, and changing a group in there will swap out the auth state in the app and update both inputs.

rafegoldberg:
how about something like this?

erunion:
I really, really, really like that. The toggle on the auth scheme entry should probably only show up if there's more than one though.

erunion:
@rafegoldberg @madisonyocum That dropdown isn't going to be present for logged out users, so we should take that into consideration.

rafegoldberg:
Styled the `<AuthBox>` per my previous comment:

<p align=center>

erunion:
@rafegoldberg We definitely shouldn't show the dropdown if they just have a single project. That looks to be the current behavior in `GroupsList`, just the demo site always uses two projects to test with. We should make sure the lack of a dropdown in the authbox still looks ok.

This all looks great, though on endpoints with only a single auth scheme, there's a scrollbar that shows up:

Screen Shot 2019-12-06 at 11 13 32 AM

rafegoldberg:
@erunion fixed the erroneous scrollbars and moved the `<GroupList>` in to a conditional on `groups.length`; this should be good to go now!

erunion:
Looks good!

PR #221, opened 2019-06-06
🐛 we want to run babel through all of the js files to check for IE compatibility

dok:
Fix is related to the ticket regarding `.startsWith`.

The fix is essentially to remove the babel exclusion for js files; This is allow babel to catch any files with compatibility issues with older browsers.

### DIFF
--- a/packages/api-explorer/package-lock.json
+++ b/packages/api-explorer/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@readme/api-explorer",
-  "version": "4.2.0",
+  "version": "4.3.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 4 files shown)


dok:
tested with IE11. It works!

@erunion I hear you're going to be taking over this repo and its deployment processes. How do you want me to notify you when something is ready to be merged and shipped?

I could tag as `ready` and assign you as a reviewer as a way

erunion:
@dok Feel free to just @, or formally assign the PR to, me whenever you're ready.

packages/api-explorer/package-lock.json
```diff
@@ -11245,41 +11306,31 @@
       }
     },
     "react-jsonschema-form": {
-      "version": "github:domharrington/react-jsonschema-form#fa1cf44205261cf5582e3c12cecf0fd39f548c44",
+      "version": "github:domharrington/react-jsonschema-form#f056a1ed34d2d0841ae897303e9d29b9e8143687",
```
kanadgupta: I believe this is an outdated version of react-jsonschema-form: https://github.com/domharrington/react-jsonschema-form/commit/f056a1ed34d2d0841ae897303e9d29b9e8143687

We should be using this one instead: https://github.com/domharrington/react-jsonschema-form/commit/fa1cf44205261cf5582e3c12cecf0fd39f548c44

erunion:
@dok Can you revert the `react-jsonschema-form` change, and resolve the conflicts with the two `package-lock.json` files?

dok:
@erunion reverted

PR #224, opened 2019-06-21
♻️ be more selective about which docs to lazy render

dok:


domharrington:
@erunion think this is good to go! @kanadgupta can you work together to build a release for this?

packages/api-explorer/package-lock.json
```diff
@@ -11320,30 +11317,20 @@
       },
       "dependencies": {
         "ajv": {
-          "version": "6.10.0",
```
erunion: @dok Did you mean to downgrade ajv here?

dok: I did not intend to downgrade. The update to `package-lock.json` was a result of running `npx lerna bootstrap`.

erunion: K cool, just checking. Looks like 5.5.2 falls in line with what the `react-jsonschema-form` dependency is requesting anyways so 👍

PR #231, opened 2019-07-24
Adding support for displaying multiple response media types on an endpoint.

erunion:
This adds support for rendering out responses that support multiple media types and is built off of the great work done in https://github.com/readmeio/api-explorer/pull/208.

### DIFF
--- a/example/swagger-files/multi-media-types.json
+++ b/example/swagger-files/multi-media-types.json
@@ -14,6 +14,29 @@
       "get": {
         "summary": "Get results",
         "description": "",
+        "requestBody": {
+          "content": {
+            "application/json": {
+              "schema": {
+                "type": "object",
+                "properties": {
+                  "string (format: date)": {
+                    "type": "string",
+                    "format": "date"
+                  },
+                  "string (format: date-time)": {
+                    "type": "string",
+                    "format": "date-time"
+                  },
+                  "string (format: string)": {
+                    "type": "string",
+                    "format": "string"
+                  }
+                }
+              }
+            }
+          }
+        },
         "responses": {
           "200": {
             "description": "OK",
@@ -80,8 +103,7 @@
                 "examples": {
                   "cat" : {
                     "summary": "An example of a cat",
-                    "value":
-                    {
+                    "value": {
                       "name": "Fluffy",
                       "petType": "Cat",
                       "color": "White",
--- a/example/swagger-files/multi-media-types.json
+++ b/example/swagger-files/multi-media-types.json
@@ -12,19 +12,40 @@
   "paths": {
     "/results": {
       "get": {
+        "summary": "Get results",
         "description": "",
         "responses": {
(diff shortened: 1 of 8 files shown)


erunion:
@rafegoldberg @madisonyocum This needs some UI love. Here are the handful of different flows that might occur with this:

With a media type that's a `text/plain` and just a string response.

Screen Shot 2019-07-24 at 4 07 20 PM

With an endpoint that doesn't have multiple media types, but does have multiple examples for what it does return. This flow is currently in production, the difference with this PR is that I'm adding the "Examples" header.

Screen Shot 2019-07-24 at 4 07 24 PM

Multiple media types, and multiple examples on the selected media type.

Screen Shot 2019-07-24 at 4 07 31 PM

Single media type, single example. No changes to this flow.

Screen Shot 2019-07-24 at 4 07 35 PM

rafegoldberg:
I think the best solutions would simply be to tighten up the display of what we've already got! something like this, maybe?

image

rafegoldberg:
per @madisonyocum's suggestion:

image

erunion:
Awesome. Let's hold off on any more changes until Marc and Dom are back from PTO so we can all sit down and chat about this and support for multiple mime type support on request bodies too. We might as well bundle it all into one large PR.

madisonyocum:
UI looks good to me!

madisonyocum:
Taking a second look at this with @erunion, the issues were copy/ proximity and changing it to 'Set an example' instead of just listing both. Also, when a user selects an example we could show a + or > of some sort, thoughts? @mjcuva https://www.figma.com/file/Jk6fzJoxVCGU3BpgP2NdHi/API-Explorer-Media-Types?node-id=3%3A9

content-type, no +
content-type
content-type-1

erunion:
Talked offline with @madisonyocum and after seeing the `content-type` dropdown idea I had in action, I think moving it down to the level of the parameters makes more sense so as to not obfuscate the code sample copy abilities of the user.

I really like the `+` idea to coorelate the media type with the example. It makes the relationship clear without being super explicit about it in the UI. Maybe a `>` or `→` instead of a `+`?

madisonyocum:
Updated it to move the content-type to the same line as the Body Param line, since the content is changing underneath and changed the + to an -> 
content-type

erunion:
Love it!

mjcuva:
Getting there! 

* I feel like the `application/json` next to `Body Params` isn’t clear what that is for. Maybe adding a `Content Type` label? 

* Can we change the `Media Type` label to `Response Type`? I really don’t like the term Media Type, since it’s not obvious what it is if you aren’t familiar with swagger

madisonyocum:
@mjcuva @erunion Updated w/ changes!
content-type-1
content-type

erunion:
Ok cool, I'm going to close this out and re-open a new PR when it's ready for review again.

PR #232, opened 2019-07-30
Adding support for $ref pointers to additionalProperties

erunion:
This incorporates a branch on our existing [react-jsonschema-form](https://github.com/mozilla-services/react-jsonschema-form) fork (now moved to our @readmeio org), to add support for `$ref` pointers on `additionalProperties` entries in data models.

Previously, if there was no `type` detected within `additionalProperties`, react-jsonschema-form would attempt to guess the type of the object as a `string`. Now, if a `$ref` pointer is present, it will be loaded as expected.

See https://github.com/readmeio/react-jsonschema-form/pull/1 for the full changeset of that fix and https://github.com/readmeio/react-jsonschema-form/pull/2 for the full changeset of what we're running in this (temporary) react-jsonschema-form fork.

### Before
62160626-49edc280-b2c9-11e9-9d63-34c995952f2e

### After
62160569-21fe5f00-b2c9-11e9-8dd5-ed7f921f24d6

### DIFF
--- a/example/swagger-files/additional-properties.json
+++ b/example/swagger-files/additional-properties.json
@@ -29,7 +29,21 @@
                     "type": "object",
                     "additionalProperties": { "type": "integer" }
                   },
-                  "object with `additionalProperties: { type: array, items: { type: integer } }` (this does not work)": {
+                  "object with `additionalProperties: { type: object, properties: ... }`": {
+                    "type": "object",
+                    "additionalProperties": {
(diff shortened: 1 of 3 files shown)


erunion:
Another good before/after with a spec that a customer is running in production:

Screen Shot 2019-07-31 at 3 32 22 PM

screen_shot_2019-07-31_at_3 22 54_pm

PR #233, opened 2019-07-30
🐛 manual endpoints may not have a URL specified in the OAS spec

dok:
Resolves a bug where projects load pages that are synced from a spec file in addition to pages that are manually added.

### DIFF
--- a/packages/api-explorer/src/lib/get-path-operation.js
+++ b/packages/api-explorer/src/lib/get-path-operation.js
@@ -1,7 +1,7 @@
 const getPath = require('./get-path');
 
 module.exports = function getPathOperation(swagger, doc) {
-  if (swagger.paths && doc.swagger) {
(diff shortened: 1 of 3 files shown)


packages/api-explorer/src/lib/get-path-operation.js
```diff
@@ -1,7 +1,7 @@
 const getPath = require('./get-path');
 
 module.exports = function getPathOperation(swagger, doc) {
-  if (swagger.paths && doc.swagger) {
+  if (swagger.paths && doc.swagger && getPath(swagger, doc)) {
```
erunion: Can you set `getPath(swagger, doc)` to a constant and use that here and the line below?

dok: Done!

PR #252, opened 2019-08-06
Updating our react-jsonschema-form fork to the latest HEAD.

erunion:
This updates our [react-jsonschema-form](https://github.com/readmeio/react-jsonschema-form) fork to the latest HEAD of our [`dist-committed`](https://github.com/readmeio/react-jsonschema-form/tree/dist-committed) branch.

Fixes that are included in this changeset are the following:

* Preventing data loss on `additionalProperties` key changes. https://github.com/readmeio/react-jsonschema-form/pull/3
  * Since I opted to not write tests on this in the library, instead doing manual testing, the way you can test this is to:
     1. Load http://localhost:9966/?selected=swagger-files%2Fadditional-properties.json
     2. Click on the "object with 'additionalProperties: { type: object, properties: ... }' and custom title"  parameter.
     3. Add a new entry. 
     4. In the `newKey` input box, type in a new value.
     5. Click out of the input. Your input data should be retained and the key+labels should change as they do with the other parameter types present on that API example.
* Backporting a previous fix that was made by @domharrington in https://github.com/mozilla-services/react-jsonschema-form/pull/1292. https://github.com/readmeio/react-jsonschema-form/pull/4

### DIFF
--- a/example/swagger-files/additional-properties.json
+++ b/example/swagger-files/additional-properties.json
@@ -43,6 +43,21 @@
                       }
                     }
                   },
+                  "object with `additionalProperties: { type: object, properties: ... }` and custom title": {
+                    "type": "object",
+                    "additionalProperties": {
+                      "type": "object",
+                      "properties": {
+                        "code": {
+                          "type": "integer"
+                        },
+                        "text": {
+                          "type": "string"
+                        }
+                      },
+                      "title": "This is a custom title"
+                    }
+                  },
                   "object with `additionalProperties: { type: array, items: { type: integer } }`": {
                     "type": "object",
                     "additionalProperties": {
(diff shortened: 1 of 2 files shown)


erunion:
There's a current bug with the dataloss fix I put in where the keychange doesn't happen for both sides of the row:

Screen Shot 2019-08-05 at 6 27 04 PM

"This is a custom title" should also read "tkktkt" since that matches the flow for `additionalProperties` schemas without a `title` defined. However, since the fix I put in place is still better than users losing their data, and this is purely a UI thing, I don't think this should hold up release.

I'm going to package this fix up into a PR for the main react-jsonschema-form repo and see if they all have any ideas on a better solution than what I've currently got.

PR #259, opened 2019-08-06
If the explorer is disabled, ArrayField components should be addable but uneditable

erunion:
| [🎟 Asana](https://app.asana.com/0/681252538274980/1132986302265166/f) |
| --- |

This resolves an unintended symptom in the explorer currently where if the explorer is disabled in the OAS via `x-explorer-enabled`, `array<object>` elements cannot be unfolded, resulting in inaccessible parameter documentation. As such:

Screen Shot 2019-08-06 at 1 37 19 PM

Instead, these schemas should be able to be seen (and added into the view), but should remain in `readOnly` mode as the rest of the explorer already is. As such:

Screen Shot 2019-08-06 at 1 38 56 PM


erunion:
@domharrington Pushed some changes up to remove the add button after the user's clicked to add an entry within the `ArrayField`.

I'm not quite sure how to test this out with enzyme though. Using `mount` and `simulate('click')` on `.field-array .array-item-add` doesn't seem to actually create any new elements on `elem`. Ideas?

erunion:
Talked offline and I'm going to try to adjust it so arrays of primitives can't be expanded.

erunion:
Arrays of primitives can no longer be expanded.

Screen Shot 2019-08-12 at 4 13 30 PM

Introduced functionality of preventing the further expansion of an `array<object>` is still in place:

Screen Shot 2019-08-12 at 4 13 36 PM

example/swagger-files/types.json
```diff
@@ -462,7 +473,7 @@
       }
     }
   },
-  "x-explorer-enabled": true,
+  "x-explorer-enabled": false,
```
domharrington: Did you mean to commit this?

erunion: lol no

PR #279, opened 2019-08-16
Moving non-HTTP code filtering down into create-docs.

erunion:
This resolves a regression that was introduced in https://github.com/readmeio/api-explorer/pull/211 where in certain cases, non-endpoint docs on ReadMe would be filtered out from being shown because they didn't have an HTTP method we recognized.

Instead, we're now doing this filtering within `create-docs` where we're parsing the OAS and know that we have an endpoint at hand.

Also added a new unit test for `create-docs` to make sure that we don't regress on the common parameter functionality introduced in #211 and begin showing stateless `parameters` endpoints within the explorer.

### DIFF
--- a/packages/api-explorer/lib/create-docs.js
+++ b/packages/api-explorer/lib/create-docs.js
@@ -41,6 +41,13 @@ module.exports = (oas, apiSetting) => {
 
       const tag = operation.tags && operation.tags.length ? operation.tags[0] : path;
       if (!docs.find(category => category.slug === tag && category.type === 'basic')) {
+        // If the method present isn't an HTTP method that we're aware of, ignore it. This can
+        // happen in the case of an API definition using something like common parameters.
+        // https://swagger.io/docs/specification/describing-parameters
+        if (!['get', 'post', 'put', 'patch', 'delete', 'head', 'options', 'trace'].includes(method)) {
+          return;
+        }
+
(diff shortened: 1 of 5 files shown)


packages/api-explorer/lib/create-docs.js
```diff
@@ -41,6 +41,15 @@ module.exports = (oas, apiSetting) => {
 
       const tag = operation.tags && operation.tags.length ? operation.tags[0] : path;
       if (!docs.find(category => category.slug === tag && category.type === 'basic')) {
+        // If the method present isn't an HTTP method that we're aware of, ignore it. This can
+        // happen in the case of an API definition using something like common parameters.
+        // https://swagger.io/docs/specification/describing-parameters
+        if (
+          !['get', 'post', 'put', 'patch', 'delete', 'head', 'options', 'trace'].includes(method)
```
domharrington: How does this work when common parameters are included via an OAS file from ReadMe? `create-docs` is only used within the demo app.

erunion: That's a good point...

I'll do some further testing and send it back for review.

domharrington: I *think* the previous approach was sound, but it was just broken in that it excluded regular pages, and not common parameter endpoints?

PR #318, opened 2019-09-25
Editable server variables

domharrington:
I decided to have a go at making server variables editable in the same manner as other properties.

image

You can see this in action here: http://localhost:9966/?selected=swagger-files%2Fserver-variables.json.

At a high level, I had to modify two main parts of the code to add support for this:

- parameters-to-json-schema - to make it aware of server variables so the form module constructs a form for it
- oas-to-har - pass the values from the form into here so we can construct a URL with the server variables the user has typed in

This isn't quite as well supported as auth, which shares the state across endpoints when it's modified in one, but maybe this is good enough for now. Eventually we'll be able to move server variables state one level up and share them, like we do currently for auth.

### DIFF
--- a/packages/api-explorer/__tests__/PathUrl.test.jsx
+++ b/packages/api-explorer/__tests__/PathUrl.test.jsx
@@ -25,9 +25,19 @@ const props = {
 test('should display the path', () => {
   const pathUrl = shallow(<PathUrl {...props} />);
 
-  expect(pathUrl.find('span.url').text()).toBe(oas.servers[0].url);
-  expect(pathUrl.find('span.api-text').text()).toBe('/pet/');
-  expect(pathUrl.find('span.api-variable').text()).toBe('petId');
+  expect(pathUrl.find('span.url').text()).toBe(oas.url());
+  expect(pathUrl.find('span.path span.api-text').text()).toBe('/pet/');
+  expect(pathUrl.find('span.path span.api-variable').text()).toBe('petId');
+});
+
+test('should display the url with variables', () => {
+  const serverVariablesOas = new Oas({
+    servers: [{ url: 'https://example.com/{path}', variables: { path: { default: 'path' } } }],
+  });
+
+  const pathUrl = shallow(<PathUrl {...props} oas={serverVariablesOas} />);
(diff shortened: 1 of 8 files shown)


domharrington:
Speaking about this over the phone, it would make more sense to do this via a popover in the same way auth is done. Then we can share the server variables across all endpoints, and when a change is made in one, it will mirror across all of them - this mirrors how auth works.

Right now auth is calculated once at the very top layer:
https://github.com/readmeio/api-explorer/blob/bd6a5fe395bd23aa9121463cadbed6d50385aff9/packages/api-explorer/src/index.jsx#L31

Since server variables will be done in much the same way, it may make sense to add another layer around that piece, like so:

```
globals: {
  auth: getAuth(this.props.variables.user, this.props.oasFiles),
  serverVariables: getServerVariables(),
},
```

And do the same for other places where auth is passed around as well:

https://github.com/readmeio/api-explorer/blob/bd6a5fe395bd23aa9121463cadbed6d50385aff9/packages/api-explorer/src/lib/oas-to-har.js#L84

```
globals = { auth: {}, serverVariables: {} },
```

See https://github.com/readmeio/api-explorer/blob/bd6a5fe395bd23aa9121463cadbed6d50385aff9/packages/api-explorer/src/index.jsx#L37-L43 for how the auth change callback is passed all the way down to the input. I can't remember specifically, but I think I had to do something

I had to do this: https://github.com/readmeio/api-explorer/blob/bd6a5fe395bd23aa9121463cadbed6d50385aff9/packages/api-explorer/src/SecurityInput.jsx#L15-L25 to make sure that the updating of the <input> wasn't blocked waiting for all of the other inputs to update as well. It seems to work pretty well.

Doing this change may also mean we can remove server variables from parameters-to-json-schema, since they're always strings anyway? And will come from the very top level?

perryloh-nintex:
@domharrington @erunion any ideas when this feature can be made available from readme.io ?

erunion:
@perryloh-nintex We don't have any timetables on getting this feature wrapped up at the moment as it's been deprioritized a bit.

erunion:
We're going to be addressing this in the redesign so I'm going to close this out.

PR #325, opened 2019-10-01
deps: Upgrade ESLint to 6.5.0

erunion:


packages/api-explorer/src/AuthBox.jsx
```diff
@@ -10,7 +10,7 @@ function Securities({ authInputRef, operation, onChange, oauth, auth, onSubmit }
     const securities = securityTypes[type];
     return (
       <form key={type} onSubmit={onSubmit}>
-        <h3>{type} Auth</h3>
+        <h3>{`${type} Auth`}</h3>
```
domharrington: Weird change...

packages/oas-extensions/package.json
```diff
@@ -13,9 +13,9 @@
   "license": "ISC",
   "repository": "https://github.com/readmeio/api-explorer/tree/master/packages/oas-extensions",
   "devDependencies": {
-    "eslint": "^5.3.0",
-    "eslint-config-airbnb-base": "^13.1.0",
-    "eslint-plugin-import": "^2.17.2",
+    "eslint": "^6.5.0",
```
domharrington: I think in lerna, they encourage moving shared devdeps up to the top level and using them from there. This may not work for executables though.

packages/api-explorer/src/AuthBox.jsx
erunion: It was complaining about this as a React object needing to be on its own line, so it was trying to convert this into:

```
<h3>
  {type}
  {' '}
  Auth
</h3>
```

Silly.

packages/oas-extensions/package.json
rafegoldberg: It should work, or at least it has for the various Babel and Webpack plugins I've tried adding in the past!

erunion: Oh good to know. I'll see if I can pull them out of here these individual packages and clean those lock files up.

erunion:
@domharrington TBH, I have had zero value from Code Climate. Most of the time the stuff it complains about are either having too many return statements within a function, or a function being longer than 25 lines and I'm like "🤷‍♂ ok?".

JSInspect and ESLint have saved more potential issues cropping up for me than anything else.

domharrington:
If you have codeclimate permissions, im okay with removing it! Same with jsinspect as well if we get most value from eslint and prettier.

erunion:
I'm having trouble getting devDependencies for ESLint in the root `package.json` to carry down into our monorepo packages so I'm going to merge this in as-is and we can revisit this cleanup later.

PR #312, opened 2019-09-17
feat: Adding support for multiple media types

erunion:
###### See https://github.com/readmeio/api-explorer/pull/231 for previous discussion.

This adds support for rendering out responses that support multiple media types and is built off of the great work done in #208.

### Mockups
63114707-8117d100-bf63-11e9-85e3-1cab0cf5899f

63114708-8117d100-bf63-11e9-9f57-a60fe622052f

PR #346, opened 2019-10-23
fix: Backport support for multi media types to legacy Manual APIs

erunion:
### Problem
When we shipped https://github.com/readmeio/api-explorer/pull/312 two weeks ago, it inadverantly broke all APIs that were manually documented through our manual editor because those APIs have their languages (media types) as a `language` property, but we're now building up (via the code shower utility) a `languages` array.

Since the explorer now expects `languages` to be present, all manually documented APIs were crashing.

### Fix
Since changing the shape of this data would require a large database migration, and we're currently in the process of redoing our manual editor to be OAS first (which will the code shower utility already handles), I've decided to write some adhoc code to upgrade these response examples on the fly to the new, expected, format.

### DIFF
--- a/packages/api-explorer/src/EndpointErrorBoundary.jsx
+++ b/packages/api-explorer/src/EndpointErrorBoundary.jsx
@@ -18,17 +18,14 @@ class EndpointErrorBoundary extends React.Component {
 
   render() {
     if (this.state.error) {
-      const mailTo = (
-        <a href="mailto:support@readme.io?subject=API Explorer Error">support@readme.io</a>
-      );
-
       return (
         <div className="hub-reference-section">
           <div className="hub-reference-left" style={{ paddingLeft: '2%' }}>
             <h3>
               There was an error rendering this endpoint. If you are the owner of this project
-              please contact
-              {` ${mailTo} with the following error:`}
+              please contact&nbsp;
+              <a href="mailto:support@readme.io?subject=API Explorer Error">support@readme.io</a>
+              &nbsp;with the following error:
             </h3>
             <BoundaryStackTrace error={this.state.error} info={this.state.info} />
           </div>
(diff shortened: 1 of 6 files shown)


packages/api-explorer/src/EndpointErrorBoundary.jsx
```diff
@@ -18,17 +18,14 @@ class EndpointErrorBoundary extends React.Component {
...
-      );
-
       return (
         <div className="hub-reference-section">
           <div className="hub-reference-left" style={{ paddingLeft: '2%' }}>
             <h3>
               There was an error rendering this endpoint. If you are the owner of this project
-              please contact
-              {` ${mailTo} with the following error:`}
+              please contact&nbsp;
```
erunion: These `mailTo` changes are because when errors were present, the copy for this header was:

> There was an error rendering this endpoint. If you are the owner of this project please contact [Object object] with the following error:

packages/api-explorer/src/ResponseExample.jsx
```diff
@@ -41,15 +42,30 @@ function getReactJson(example, responseExampleCopy) {
...
+) {
   const mediaTypes = example.languages;
   if (mediaTypes.length <= 1) return false;
 
   let responseMediaTypeCopy = responseMediaType;
   if (!responseMediaTypeCopy && mediaTypes[0]) responseMediaTypeCopy = mediaTypes[0].languages;
 
   return (
-    <div>
+    <div className={nestInTabberBarDivClass ? 'tabber-bar' : ''}>
```
erunion: This change is in here because with some CSS that we added for the multiple media type dropdown adds top-padding if `tabber-bar` is present. Problem is that if you don't have multiple media types to show, we were still showing that padding because the `<div class="tabber-bar"></div>` element was present and empty. 

Now if we aren't going to be showing a dropdown, we won't add the `tabber-bar` div into the DOM.

rafegoldberg: Ohh word, good catch. I think I tried to solve this with a fancy-ass `:empty` selector, but ran in to some road blocks around the markup structure! This looks like a good solution.

PR #348, opened 2019-10-24
Fix/syntax highlighter

dok:
A value of `false` is being passed down into the API explorer via code-examples. We need to make sure that CodeMirror never is passed down anything other than a string.


dok:
@domharrington that's a great point. I'll make sure it always casts to string

PR #364, opened 2019-11-06
Fix issue where responses stored as a $ref wouldn't be fully loaded

erunion:
When a response is stored within a `$ref` component, the explorer wouldn't display it in response tab window, and the description (if present) would be missing from the response schema body box.

Screen Shot 2019-11-05 at 4 48 12 PM

With this fix in place, `$ref` responses appear as they would if they weren't loaded as a component:

Screen Shot 2019-11-05 at 4 50 01 PM

### DIFF
--- a/packages/api-explorer/src/ResponseSchema.jsx
+++ b/packages/api-explorer/src/ResponseSchema.jsx
@@ -87,6 +87,14 @@ class ResponseSchema extends React.Component {
     const { operation, oas } = this.props;
     if (!operation.responses || Object.keys(operation.responses).length === 0) return null;
     const schema = this.getSchema(operation);
+
+    let response = operation.responses[this.state.selectedStatus];
+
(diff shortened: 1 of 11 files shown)


packages/api-explorer/src/ResponseSchema.jsx
```diff
@@ -87,6 +87,14 @@ class ResponseSchema extends React.Component {
     const { operation, oas } = this.props;
     if (!operation.responses || Object.keys(operation.responses).length === 0) return null;
     const schema = this.getSchema(operation);
+
+    let response = operation.responses[this.state.selectedStatus];
+
+    // @todo This should really be called higher up when the OAS is processed within the Doc component.
+    if (response.$ref) {
```
erunion: I don't particularly like doing this schema definition lookup here and in `create-code-shower`, but moving it further up the stack is going to require a heavy refactor.

PR #376, opened 2019-11-12
Extracting oas-to-har into @readme/oas-to-har

erunion:
* [x] Extracting [`oas-to-har.js`](https://github.com/readmeio/api-explorer/blob/master/packages/api-explorer/src/lib/oas-to-har.js) into a new @readme/oas-to-har package.
* [x] Replacing usages of [`Oas`](https://github.com/readmeio/api-explorer/blob/master/packages/api-explorer/src/lib/Oas.js) with the improved https://github.com/readmeio/oas module. See https://github.com/readmeio/oas/pull/62 for more information.
* [x] Adopting https://github.com/readmeio/eslint-config in the api-explorer package.


packages/api-explorer/src/Doc.jsx
```diff
@@ -16,9 +18,7 @@ const Response = require('./Response');
 const ResponseSchema = require('./ResponseSchema');
 const EndpointErrorBoundary = require('./EndpointErrorBoundary');
 
-const Oas = require('./lib/Oas');
-const { Operation } = require('./lib/Oas');
-const getPath = require('./lib/get-path');
```
rafegoldberg: Did you mean to delete this dependency? I see that `getPath` is still being referenced further down in this file... haven't fully traced so I very well may be wrong!

erunion: Yep, `get-path` now exists in https://github.com/readmeio/oas and can be imported via `require('oas/utils').getPath`.

erunion: See https://github.com/readmeio/oas/blob/master/utils.js

rafegoldberg: perfect!

rafegoldberg: merge away

Issue #262, opened 2019-08-08
Should we surface string formats?


erunion:
Currently if a `string` has a special format like `date`, `url`, `binary`, etc we only render it in the UI as a `string`. 

![Screen Shot 2019-08-07 at 5 00 58 PM](https://user-images.githubusercontent.com/33762/62665648-f7e61600-b934-11e9-948d-97313e6661fc.png)

It might give some more clarity to the person reading the docs what kind of string they should be supplying (especially if an example or default isn't being documented by the API owner).

Only real downside or hurdle I see to doing this is that our current css in `api-explorer.css` will need a major rewrite to support something this dynamic. Or maybe we should take this as an opportunity to redesign how we represent those formats entirely?

domharrington:
Would it be possible for us to do this in CSS? Similar to how we do the number formats?

https://github.com/readmeio/api-explorer/blob/70909b5ac656d367425bbc208296fede5a2f1188/packages/api-explorer/api-explorer.css#L424

https://github.com/readmeio/api-explorer/blob/70909b5ac656d367425bbc208296fede5a2f1188/packages/api-explorer/src/form-components/SchemaField.jsx#L60

rafegoldberg:
@domharrington that would work, but wouldn't you have to manually upkeep the styles/classes for every variation of type combinations?

erunion:
We don't currently keep the `format` anywhere in the view, so it would require a little bit of work to `SchemaField`, but yeah I'm mostly concerned about creating another large list of CSS overrides to maintain like we already have in there.

domharrington:
Yeah agree. I'm open to other suggestions for sure but that seemed like the quickest win for now!

PR #228, opened 2019-07-16
Update Schema Form UI

rafegoldberg:
|🎟 [Asana](https://app.asana.com/0/681252538274980/1125996951109079/f)|🛑 [Broken](https://developer.juspay.in/reference#enable-mandate-feature-in-order)|✅ [Working](https://juspay.readme-stage-pr-1746.readme.ninja/reference#enable-mandate-feature-in-order)|
|:-:|:-:|:-:|

[The `react-jsonschema-form` plugin](https://www.npmjs.com/package/) we're using is a pain in the ass to style with pure CSS. Instead, I opted to write [a custom template](https://react-jsonschema-form.readthedocs.io/en/latest/advanced-customization/#field-template), which let me move the description text _under_ the key and type labels:

image

This frees up more horizontal space and should solve the overflow issues. I've also touched up the styling of object and array parameters to make them more visually coherent:

image

I still have some questions re: release process for this repo but please pull down this branch and [take a peek at the changes locally](http://localhost:9966/?selected=swagger-files%2Ftypes.json). Let me know if all looks good and I'll figure it out from there!

PR #381, opened 2019-11-22
Adding support for more string formats

erunion:
This adds support for a handful of new string `format` types:

* `date-time` (`dateTime` was already supported, but `date-time` is an alternative to this)
* `password`
* `timestamp`
* `uri`

These types are now represented in the format decorators next to the schema field, as such:

Screen Shot 2019-11-22 at 8 51 29 AM

Additionally, the `input` elements for `date-time` and `password` formats are now contextual to the format type. `date-time` (and `dateTime`) have date selectors, and text in a `password` field is hidden.

Screen Shot 2019-11-22 at 8 53 32 AM

I've also taken some time to clean up our `Params` component as it was beginning to get a bit messy.

Resolves https://github.com/readmeio/api-explorer/issues/262


packages/api-explorer/style/main.scss
```diff
@@ -286,235 +286,6 @@ form.rjsf button[disabled] {
   color: #d2d2d2 !important;
 }
 
-/* Add type labels. */
```
erunion: None of this junk is used anymore since @rafegoldberg did https://github.com/readmeio/api-explorer/pull/228 🥇

rafegoldberg:
This looks good to me!

PR #404, opened 2019-12-12
Fixing a regression of additionalProperties keys not being editable

erunion:
This resolves a regression of functional on schemas with `additionalProperties` no longer having editable labels/keys.
 
### Broken
Screen Shot 2019-12-12 at 3 19 51 PM

### Resolved
Screen Shot 2019-12-12 at 3 19 47 PM

*Note:* Keys don't update until you focus out of the input, but this is the same behavior that RJSF has currently https://rjsf-team.github.io/react-jsonschema-form/

### DIFF
--- a/packages/api-explorer/src/form-components/SchemaField.jsx
+++ b/packages/api-explorer/src/form-components/SchemaField.jsx
@@ -11,6 +11,7 @@ const React = require('react');
 const PropTypes = require('prop-types');
 
 const BaseSchemaField = require('react-jsonschema-form/lib/components/fields/SchemaField').default;
+const { ADDITIONAL_PROPERTY_FLAG } = require('react-jsonschema-form/lib/utils');
 
 function getDefaultNumFormat(type) {
   if (type === 'integer') return 'int32';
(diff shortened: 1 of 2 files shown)


packages/api-explorer/src/form-components/SchemaField.jsx
```diff
@@ -70,13 +71,36 @@ function getTypeLabel(schema) {
...
+  if (ADDITIONAL_PROPERTY_FLAG in schema) {
+    editableLabel = (
+      <input
+        defaultValue={label}
+        id={`${id}-key`}
+        onBlur={event => onKeyChange(event.target.value)}
+        type="text"
+      />
+    );
+  }
```
rafegoldberg: Can we refactor this to move the conditional in to the JSX template?

```jsx
<label className="label-name" htmlFor={id}>
  {ADDITIONAL_PROPERTY_FLAG in schema ? EditLabel : label}
  // etc.
</label>
```

That way we can simplify the preceding `editableLabel` snippet:

```js
const EditLabel = <input
  defaultValue={label}
  id={`${id}-key`}
  onBlur={event => onKeyChange(event.target.value)}
  type="text"
/>;
```

rafegoldberg: (obviously not a functional concern; don't hold your merge over it or anything!)

PR #431, opened 2020-01-09
Align Key/Val Param Fields

rafegoldberg:
|🎟 [Asana](https://app.asana.com/0/681252538274980/1153860678793593/f)|
|:--:|

Some key/value fields do not align very well when nested. 

**Broken**: go to [this example page](https://stackpath.dev/reference/workloads#updateworkload) and scroll down to Annotations (nested under Metadata) and click the Add button to create a new set of key/value pairs.


erunion:
Build failed because I killed our CodeClimate account. Removed it here: https://github.com/readmeio/api-explorer/pull/432

PR #435, opened 2020-01-10
chore: update beta/editor-rebase with latest master

erunion:


erunion:
@rafegoldberg There are some Markdown tests broken here, but I'm not sure if they were already broken, or broke with this merge.

rafegoldberg:
# 🙌

Dang, that was quick! And yeah, those tests were broken to begin with (the only thing holding us back from launch...) Will look this over more specifically this weekend. On that note, I'd love to grab some time with you next week to go over these tests and make surer e'erything is covered appropriately, since testing is _not_ my strong suit!

PR #436, opened 2020-01-11
Support for allOf and a host of small improvements and bug fixes

erunion:
## 🌱 Improvements
* [x] Move [react-jsonschema-form](http://npm.im/react-jsonschema-form) off our outdated fork and into a new soft fork at `@readme/react-jsonschema-form`, which is running the latest version of the upstream project.
* [x] ~Backporting https://github.com/rjsf-team/react-jsonschema-form/pull/954 back into our fork until a future release of RJSF will support it directly.~
  * This work is now directly in RJSF and our fork so we no longer need any special handling for component schemas.
* [x] Adding support for `format: blob` and `format: html` on string params to show a textarea.
* [x] Support `allOf` in the form.
* [x] Implementing a hot fix for some `oneOf` and `anyOf` UI quirks. See https://github.com/readmeio/api-explorer/pull/436#issuecomment-597428988 for details.
* [x] Adding support for `allOf`, `oneOf`, and `anyOf` in response schemas.
  * For `oneOf` and `anyOf` cases, since we can't really flatten down every option defined in those arrays into a single schema that would validate against the document, I'm choosing to just select the first defined object in the set and flatten that out.
* [x] Upgrading [@readme/oas-tooling](https://github.com/readmeio/oas) to [3.0.0](https://github.com/readmeio/oas/blob/master/CHANGELOG.md#v300)
  * Brings enhancements to handle polymorphism cases in our response schema rendering
  * Fixes a [bug](https://app.asana.com/0/681252538274980/1164563834588954/f) where we couldn't handle escaped schema definitions.
  * Updates our JSON Schema generation to function on the newer release of RJSF.
* [x] Improved error messaging states and the ability for us to capture errors for reporting. https://github.com/readmeio/api-explorer/pull/517

## 🐛 Bug fixes
* [x] Adding support for being able to handle `$ref` pointers that are escaped.
  * For example, previously if a ref was `Pet~1Error`, the actual schema we should look for is `Pet/Error` but we'd look for the `Pet~1Error` instead. We now properly handle these escaped cases.
  * [swagger2openapi](https://www.npmjs.com/package/swagger2openapi) can't currently handle these use cases, but I've submitted a PR to that repo to resolve this. https://github.com/erunion/oas-kit/commit/0d8145318424d66aa186d5aa650651f3e6874983
* [x] Adding word wrapping on response schema bodies so long schema names won't get hidden in the overflow.
* [x] Fixing a bug where `oneOf` schemas nested inside of an object that were populated via a `$ref` would fail. Fixes https://github.com/readmeio/api-explorer/issues/495.

## 👨‍🎨 Comparisons
* allOf support
  * Broken: https://preview.readme.io/?selected=swagger-files%2Fall-of.json
  * Fixed: https://api-explorer-feat-mainl-rcvkhk.herokuapp.com/?selected=swagger-files%2Fpolymorphism.json
* Error states
  * Yikes: https://preview.readme.io/?selected=swagger-files%2Fendpoint-error-boundary.json
  * Oooooo: https://api-explorer-feat-mainl-rcvkhk.herokuapp.com/?selected=swagger-files%2Fendpoint-error-boundary.json

## 🚧 Known Quirks
* [ ] Swapping between a top-level `oneOf` state preserves any data in the code sample/request that was set on the other object.
  * ❗️This is a bug in RJSF and out of scope for this PR.
* [ ] ~~Loading a `oneOf` that contains an `allOf` causes the form to appear without CSS, but toggling between the dropdown eventually fixes it. Unfortunately toggling another item in the form will break the CSS again.~~
  * ❗️There's still some remaining UI quirks with `oneOf` elements, but see https://github.com/readmeio/api-explorer/pull/436#issuecomment-597428988 for details.


domharrington:
Love the patch approach to getting off of our fork. If there's not a unit test for this case working in our code, it would be great to get one added to ensure this never regresses.

erunion:
## 🚧 Current Issues
* [ ] Swapping between a top-level oneOf state preserves any data in the code sample/request that was set on the other object.
* [ ] Loading a oneOf that contains an allOf causes the form to appear without CSS, but toggling between the dropdown eventually fixes it.

Screen Shot 2020-02-14 at 5 27 57 PM

* [ ] allOf, oneOf, anyOf in response schemas doesn't work at all right now, causing that data to just not show up.

erunion:
> Loading a `oneOf` that contains an `allOf` causes the form to appear without CSS, but toggling between the dropdown eventually fixes it. Unfortunately toggling another item in the form will break the CSS again.
>
> 

I've kind of been able to resolve this by creating a new "shell" `CustomTemplate` for these kinds of scenarios, but the children within it don't take up the full width of the layout. 

Screen Shot 2020-03-10 at 8 39 35 PM

It's not ideal to have it disjointed like this, but the CSS at least looks consistent and no longer drops in and out depending on how many times you toggle the select input. That said, since fixing this is going to be a bit more of a project, and this PR's goal is to get allOf working, I'm punting this specific UI bug for now and [filed an issue in Asana](https://app.asana.com/0/681252538274980/1165921409642121/f) to tackle it later.

example/src/Demo.jsx
```diff
@@ -14,58 +14,101 @@ require('../../example/swagger-files/response-schemas.json');
 
 require('../../packages/api-logs/main.css');
 
-function Demo({ fetchSwagger, status, docs, oas, oauth }) {
-  return (
-    <div>
-      <div className="api-list-header">
-        <ApiList fetchSwagger={fetchSwagger} />
-        <pre>{status.join('\n')}</pre>
+class Demo extends React.Component {
```
domharrington: You could have used the `useState()` hook for this if you wanted to keep it as a functional component. https://reactjs.org/docs/hooks-state.html I'm completely fine with this approach though.

example/swagger-files/one-of.json
```diff
@@ -1,156 +0,0 @@
-{
```
domharrington: Did this get moved somewhere else? Is this now in oas-tooling?

packages/api-explorer/__tests__/Params.test.jsx
```diff
@@ -159,21 +162,54 @@ test('additionalProperties object labels (keys) should be editable', () => {
...
+    );
+
+    expect(params.html()).not.toMatch('Unknown field type undefined');
+  });
 });
 
 describe('schema format handling', () => {
   describe('string types', () => {
-    it('json should render as <textarea>', () => {
+    it.each([
```
domharrington: Woah this is great!

packages/api-explorer/src/Doc.jsx
```diff
@@ -393,10 +393,12 @@ Doc.propTypes = {
   language: PropTypes.string.isRequired,
   lazy: PropTypes.bool,
   Logs: PropTypes.func,
+  maskErrorMessages: PropTypes.bool,
```
domharrington: How does this property get sent down from ReadMe?

packages/api-explorer/src/Params.jsx
```diff
@@ -37,6 +40,10 @@ function Params({
 }) {
   const jsonSchema = parametersToJsonSchema(operation, oas);
 
+  // If this operation doesn't have a set operationID, generate a random string to be one so we have unique form IDs
+  // across the explorer.
+  const operationId = 'operationId' in operation ? operation.operationId : shortid.generate();
```
domharrington: Can we make this stable, so it generates an id from a hash of the operation or something? It would break deep linking if it changed on each page refresh.

packages/api-explorer/src/Params.jsx
```diff
@@ -48,31 +55,35 @@ function Params({
...
           fields={{
-            DescriptionField,
             ArrayField,
+            DescriptionField,
             SchemaField,
           }}
           formData={formData[schema.type]}
-          id={`form-${operation.operationId}`}
-          idPrefix={operation.operationId}
+          id={`form-${schema.type}-${operationId}`}
```
domharrington: Deploying this will break existing links that may be currently shared around the web. Can we add a dummy div with the old id format so it is backwards compatible?

packages/api-explorer/style/ErrorBoundary.scss
```diff
@@ -0,0 +1,35 @@
+.hub-reference-error {
```
domharrington: cc @rafegoldberg can you take a look through this?

packages/api-explorer/__tests__/PathUrl.test.jsx
```diff
@@ -18,6 +18,7 @@ const props = {
   oas,
   oauth: false,
   onChange: () => {},
+  onGroupChange: () => {},
```
domharrington: What does onGroupChange do?

erunion:
@dom Here's a side-by-side!

Broken: https://preview.readme.io/?selected=swagger-files%2Fall-of.json
Fixed: https://api-explorer-feat-mainl-rcvkhk.herokuapp.com/?selected=swagger-files%2Fpolymorphism.json

packages/api-explorer/style/ErrorBoundary.scss
erunion: @rafegoldberg wrote this lol

packages/api-explorer/src/Doc.jsx
erunion: Haven't done that part yet, but the idea was to expose the `isAdmin` boolean from hub2 users to a `<script data-json="..." />` tag  our API Reference template in the same way we populate the rest of the Explorer.

packages/api-explorer/style/ErrorBoundary.scss
rafegoldberg: haha, but come to mention it... I should autoprefix this manually (since we've only added PostCSS to readme proper, as far as I know...) Pushing an update now.

packages/api-explorer/__tests__/PathUrl.test.jsx
erunion: That gets called when you toggle between auth groups in the auth modal. Maybe it should be renamed to `onAuthGroupChange`.

packages/api-explorer/src/Params.jsx
erunion: Re-added as a `<div>` that's wrapping the entire component!

example/src/Demo.jsx
erunion: Ended up reshuffling this addition around a bit so we can easily add new "experiment" toggles. Going to keep it as-is since I don't think setting up hooks dynamically will work with this kind of set up. Happy to revisit later as a foray into hooks though!

PR #460, opened 2020-01-27
🐛 fix broken api log url

dok:
From this thread: https://readmeio.slack.com/archives/CRE57G3J7/p1579654198009000

Fixes the logs URL. See here for another reference: https://github.com/readmeio/api-explorer/blob/master/packages/api-logs/index.jsx#L228

We currently have a temporary fix via redirects in the parent project settings


erunion:
Will merge and tag a release when it passes CI

PR #463, opened 2020-01-31
fix: stop mutating the supplied pathOperation object in oas-to-har

erunion:
This resolves a major bug in the `oas-to-har` package with regard to how we were handling common parameters. 

When we call the the package to create a HAR object to call, we'd supply it with a `pathOperation` object that represented the current path+method we're processing in the current OAS file. The problem is that inside `oas-to-har`, when constructing parameters to add into the HAR, we were doing so on the `pathOperation` object that was supplied instead of against an instance variable to that function.

This meant that every time a user would click "Try It" to make an API call, we'd re-add the same common parameters to the instanced `operation` variable further up the stack in `Doc.jsx`. This would then cascade back down into `oas-to-har` and we'd slowly begin to exponentially double up on common parameters.

🙃 

Also in this PR was another bug I uncovered while investigating this one in that when we were parsing an operation, if a `servers` object was present at the HTTP method level in the operation, we'd interpret that as an HTTP method and build out an explorer document for that. We're now hardcoding to only rendering out API operations that implement every method listed on https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods.

### DIFF
--- a/packages/oas-to-har/src/index.js
+++ b/packages/oas-to-har/src/index.js
@@ -88,40 +88,39 @@ module.exports = (
     har.url = `https://try.readme.io/${har.url}`;
   }
 
-  if (!pathOperation.parameters) {
-    // eslint-disable-next-line no-param-reassign
-    pathOperation.parameters = [];
+  // Does this operation have any parameters?
+  const parameters = [];
+  if (pathOperation.parameters) {
+    pathOperation.parameters.forEach(param => {
+      if (param.$ref) {
+        parameters.push(findSchemaDefinition(param.$ref, oas));
+      } else {
+        parameters.push(param);
+      }
+    });
   }
 
   // Does this operation have any common parameters?
   if (oas.paths && oas.paths[pathOperation.path] && oas.paths[pathOperation.path].parameters) {
     oas.paths[pathOperation.path].parameters.forEach(param => {
-      pathOperation.parameters.push(param);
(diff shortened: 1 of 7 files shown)


packages/oas-to-har/src/index.js
```diff
@@ -88,40 +88,39 @@ module.exports = (
     har.url = `https://try.readme.io/${har.url}`;
   }
 
-  if (!pathOperation.parameters) {
-    // eslint-disable-next-line no-param-reassign
-    pathOperation.parameters = [];
+  // Does this operation have any parameters?
+  const parameters = [];
```
domharrington: So this means we're creating a fresh array of parameters each time this is called? This is good for now, but eventually it would be nice to wrap this logic into here as operation.getParameters() or something, then it could be cached in there and not built out every time: https://github.com/readmeio/oas/blob/3e1983fb7b3bf04c1ba0f9286db6e0c158af8314/src/oas.js#L7

erunion: Yeah that's what'll happen for now. I'd like to at some point refactor most of this HAR library since we're doing some similar stuff throughout it: https://github.com/readmeio/api-explorer/blob/fix/dupe-request-headers/packages/oas-to-har/src/index.js#L123

Could easily replace that with a `operation.getQueryParameters()` function call.

PR #471, opened 2020-02-05
feat: setting a custom header for all "Try It" requests

erunion:
I'm iffy on the header name, but this will allow people to split apart "Try It" requests in their Developer Metrics.

### DIFF
--- a/packages/api-explorer/src/Doc.jsx
+++ b/packages/api-explorer/src/Doc.jsx
@@ -70,7 +70,7 @@ class Doc extends React.Component {
       proxyUrl: true,
     });
 
-    return fetchHar(har).then(async res => {
+    return fetchHar(har, 'ReadMe API Explorer').then(async res => {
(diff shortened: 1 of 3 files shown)


packages/api-explorer/src/Doc.jsx
```diff
@@ -70,7 +70,7 @@ class Doc extends React.Component {
       proxyUrl: true,
     });
 
-    return fetchHar(har).then(async res => {
+    return fetchHar(har, 'ReadMe API Explorer').then(async res => {
```
domharrington: Can you make this pull in the current api explorer version from the package.json? Or maybe that doesnt matter at all actually, since we'll always know the version in production is latest.

erunion: Yeah I'll do that since even though the version in production will always be the latest, it won't be so for old logs.

PR #473, opened 2020-02-09
feat: importing @readme/emojis in the markdown processor

erunion:


erunion:
merging this in because it's passing tests!

PR #493, opened 2020-02-25
fix: downgrading datetime inputs to simple text inputs

erunion:
There appears to be a bug in RJSF right now where it's disabling input into `<input type="datetime-local">` inputs in Firefox. As a quick hotfix, we're downgrading our format support of `datetime` and `date-time` to simple text inputs.

### DIFF
--- a/packages/api-explorer/src/ResponseExample.jsx
+++ b/packages/api-explorer/src/ResponseExample.jsx
@@ -249,7 +249,9 @@ class ResponseExample extends React.Component {
                         <div className="example example_json">{transformExampleIntoReactJson(example)}</div>
                       ) : (
                         // json + multiple examples is already handled in `showExamples`.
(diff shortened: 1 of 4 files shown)


packages/api-explorer/src/ResponseExample.jsx
```diff
@@ -249,7 +249,9 @@ class ResponseExample extends React.Component {
                         <div className="example example_json">{transformExampleIntoReactJson(example)}</div>
                       ) : (
                         // json + multiple examples is already handled in `showExamples`.
-                        <>{isJson && example.multipleExamples ? null : getHighlightedExample(example)}</>
+                        <React.Fragment>
```
erunion: Made this change because jsinspect couldn't handle parsing `<>` fragments.

PR #508, opened 2020-03-06
fix(markdown@v5): fix magic block null cases.

rafegoldberg:
- [x] empty Heading blocks
- [x] empty Image blocks
- [x] empty Embed blocks
- [x] non-standard Callout themes
- [x] blank Table header cells

### DIFF
--- a/packages/markdown/processor/parse/magic-block-parser.js
+++ b/packages/markdown/processor/parse/magic-block-parser.js
@@ -133,18 +133,16 @@ function tokenize(eat, value) {
             type: row ? 'tableCell' : 'tableHead',
             children: this.tokenizeInline(val, eat.now()),
           };
+          // convert falsey values to empty strings
+          sum[row].children = [...sum[row].children].map(v => v || '');
           return sum;
         }, []);
-      return eat(match)(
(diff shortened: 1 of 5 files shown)


packages/markdown/processor/parse/magic-block-parser.js
```diff
@@ -57,40 +59,40 @@ function tokenize(eat, value) {
...
     case 'callout': {
-      json.type = {
+      const types = {
         info: ['ℹ', 'info'],
         success: ['👍', 'okay'],
         warning: ['⚠️', 'warn'],
         danger: ['❗️', 'error'],
-      }[json.type];
+      };
+      json.type = json.type in types ? types[json.type] : [json.icon || '👍', json.type];
```
erunion: Is this to handle cases where people have emoji in their docs that we don't recognize as types?

rafegoldberg: Kind of, but it’s really more to account for cases in which users have hardcoded an `item` option in the magic block's JSON which doesn't correspond to one of our pre-approved themes. (Found this edge case in the [Branch docs](https://help.branch.io/).)

PR #513, opened 2020-03-09
Configure Heroku review apps pipeline.

rafegoldberg:
|🧐 [API Explorer Review App](https://api-explorer-enhancemen-thtf9h.herokuapp.com/)|
|--|

Configure a Heroku review apps pipeline for the API Explorer.

- [x] add a Procfile (1041a3e)
- [x] set up the [Heroku pipeline](https://dashboard.heroku.com/pipelines/d696b296-6ede-467e-a2f2-6303f2d805c8)

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,0 +1,1 @@
+web: npx http-server example/dist --port $PORT
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "private": true,
   "description": "Open source components that make up ReadMe's API explorer",
   "scripts": {
-    "heroku-postbuild": "npm run bootstrap && npm run build",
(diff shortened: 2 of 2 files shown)


erunion:
Damn this is rad. Is it possible at all to get the PR app to show up in the checks list on the PR?

Screen Shot 2020-03-09 at 4 49 17 PM

RJSF has this set up with Netlify and it's really nice having a link to the app there:

Screen Shot 2020-03-09 at 4 50 14 PM

erunion:
Ah I just noticed that it shows up in the activity feed:

Screen Shot 2020-03-09 at 4 53 40 PM

Maybe that's good enough.

rafegoldberg:
@erunion yeah, that was my final call in the case of the UI as well... it's not perfect, but it definitely does the trick for such minimal effort!

PR #517, opened 2020-03-11
feat: improved error messaging functionality

erunion:
## 🌱 Changes
* [x] Adds a new `maskErrorMessages` flag on the `Doc` component that will let us toggle error messages between a consumer-friendly version, and one more targeted to ReadMe project owners.
* [x] Introduces a new `onError` handler to `Doc` that will get run whenever an error happens within the Explorer. We can use this to hook it up Explorer error reporting to Sentry! Currently `onError` is expected to return an event identifier (like the README-G ids we have in Sentry) which it'll then use to display that as an error code in the frontend for project owners if they wish to contact us.
* [x] Added a new checkbox to the demo build that will allow us to easily cycle through these different messaging states.

## 🎨 Designs
Designs can be found in Figma under "Amy 20% Project".

## ⚙ Testing
You can test the error state out on http://localhost:9966/?selected=swagger-files%2Fendpoint-error-boundary.json

To cycle through the different masked states click on the "Mask error message" checkbox:

Screen Shot 2020-03-11 at 3 58 52 PM

To trigger the complete explorer failure state, click "Show fully broken state":

Screen Shot 2020-03-11 at 3 59 27 PM


erunion:
@rafegoldberg For the consumer view, I talked with @amyshi188 and it might be best if the error message extends the full width of the box since there's no error code.

rafegoldberg:
Added styling:

- **User View**:
  image
- **Admin View**: (with error code)
  image

erunion:
Migrated the new error capturing work up to our complete API Explorer failure handling code:

Screen Shot 2020-03-11 at 3 53 26 PM

Screen Shot 2020-03-11 at 3 53 29 PM

These are the messages that'll appear on projects where the explorer is **completely** broken.

PR #540, opened 2020-03-17
Bump `rehype` plugins.

rafegoldberg:
Pulls in a few Dependbot PRs and various adjustments accordingly wrt/the new markdown processor:

- [x] Update `React.Fragment` wrappers
- [x] Fix failing unit tests
- [x] Bump `rehype-remark` (closes #528)
- [x] Bump `rehype-sanitize` (closes #529)
- [x] Bump `rehype-react` (closes #530)


rafegoldberg:
So after bumping these dependencies I've got e'erything fixed up and tested in the new processor. But the _old_ processor is still failing! Fuaasdivnmlxck... Do you have any thoughts on how to proceed @erunion? Maybe we can lock the old processor's dependencies to a specific version so we don't have to update these tests? (Not sure if that'll work in a Lerna monorepo though...)

rafegoldberg:
@erunion fixed this by dropping the `markdown-magic` subpackage dependencies back to their previous version locks; simple enough... Should be good to go, pending your review!

PR #592, opened 2020-04-11
Rdmd Feedback Fixes

rafegoldberg:
## 🧰 What's being changed?

This contains what is very likely only round one of fixes in response to feedback we've  received thus far. (See also: readmeio/readme#2483) 

- [x] add a flag for "unsafe" mode for customers with complex, legacy HTML widgets
- [x] loosen compact header notation regex (hopefully this wont break everything else)
- [x] "escape" Angular `{{` interpolation in code snippets. (This fix is hacky af...)
- [x] move newline options and unsafe mode in to RDMD settings toggle


packages/markdown/index.js
```diff
@@ -81,7 +107,7 @@ export function normalize(blocks) {
     .replace(/\[block:/g, '\n\n[block:')
     .replace(/\[\/block\]/g, '[/block]\n')
     .trim()
-    .replace(/^(#+)([\w\s\d]+\n\n)/gm, '$1 $2');
+    .replace(/^(#+)((?:\w|\d)[^{#\n]+\n{1,})/gm, '$1 $2');
```
erunion, line 110: 😬

packages/syntax-highlighter/codemirror.jsx
```diff
@@ -19,6 +19,7 @@ require('codemirror/mode/python/python');
 require('codemirror/mode/ruby/ruby');
 require('codemirror/mode/shell/shell');
 require('codemirror/mode/swift/swift');
+require('codemirror/mode/sql/sql');
```
erunion, line 22:
```suggestion
require('codemirror/mode/sql/sql');
require('codemirror/mode/swift/swift');
```

erunion:
Will merge and tag a release after the builds pass

rafegoldberg:
@erunion on it

packages/markdown/index.js
rafegoldberg: _Oh_ yeah, I know... but I honestly don't know how else to address these edge-case situations? I've been trying to keep all my shittiest decisions in this normalization function so at least we can trace 'em down.

erunion: Haha, I didn't mean that as a slam. If it works it works!

PR #598, opened 2020-04-13
fix: continuous reference header display

gratcliff:
## 🧰 What's being changed?
[Asana](https://app.asana.com/0/1169848852528794/1168701716881185)
Removed target code. This is in relation to potential issues coming with reference anchors. Instead of trying to optimize and prerender to make these work, we're removing them altogether. 

This removal should only be the case for continuous reference.

## 🧪 Testing
Add a header to a continuous reference section. In the hub, you should not be able to access it via the `secion-` `href`.


erunion:
This mean that if you have continuous reference docs, you can no longer link directly to a specific operation?

gratcliff:
> This mean that if you have continuous reference docs, you can no longer link directly to a specific operation?

You can link to an operation via the slug, but not to a header with an anchor.

erunion:
@rafegoldberg Can probably answer this, but does the anchor get added when the new Markdown engine is in-use? `ApiHeader` and the `Loop` function in `Content` are only used on the legacy engine.

rafegoldberg:
@erunion correct, anchors are [generated in the component](https://github.com/readmeio/api-explorer/blob/c40a910cd25edc192c972ce8b020600a467f701e/packages/markdown/components/Heading/index.jsx#L5-L26) at the moment (though @gratcliff and I have discussed moving it in to the Unified parser itself.)

https://github.com/readmeio/api-explorer/blob/c40a910cd25edc192c972ce8b020600a467f701e/packages/markdown/components/Heading/index.jsx#L5-L26

gratcliff:
> @erunion correct, anchors are [generated in the component](https://github.com/readmeio/api-explorer/blob/c40a910cd25edc192c972ce8b020600a467f701e/packages/markdown/components/Heading/index.jsx#L5-L26) at the moment (though @gratcliff and I have discussed moving it in to the Unified parser itself.)
> 
> https://github.com/readmeio/api-explorer/blob/c40a910cd25edc192c972ce8b020600a467f701e/packages/markdown/components/Heading/index.jsx#L5-L26

@rafegoldberg @erunion wouldn't they still be added with the new engine? https://github.com/readmeio/api-explorer/blob/master/packages/api-explorer/src/index.jsx#L178 Seems to wrap these components regardless.

rafegoldberg:
@gratcliff those are just React context providers to set up the various project/user datas.

erunion:
@rafegoldberg So is there anything we need do there for this?

gratcliff:
> @gratcliff those are just React context providers to set up the various project/user datas.

Ah I see.

rafegoldberg:
One sec; to really be sure I'd have to run this locally...

rafegoldberg:
Okay, struggling to get it `link`ed up alongside the markdown processor, but I reread through the code in more depth and I'm pretty sure that the `block-type/` and loop components don't have anything to do with how the actual section header is rendered for each endpoint in a reference section, continuous or otherwise? I believe @erunion is correct that these were simply used to render magic block markdown, and will be entirely replaced by the new engine. 

This means anchors will continue to be generated for any H1-6 tag in the body of the document. Which is what you were trying to avoid for continuous reference sections, if I understand correctly @gratcliff?

rafegoldberg:
In that case, we'd need to pass [a new option](https://github.com/readmeio/api-explorer/blob/c40a910cd25edc192c972ce8b020600a467f701e/packages/markdown/options.json) to the markdown engine itself to tell it whether or not to generate anchor headings.

gratcliff:
> Okay, struggling to get it `link`ed up locally alongside the markdown processor, but I reread through the code in more depth and I'm pretty sure that the `block-type/` and loop components don't have anything to do with how the actual section header is rendered for each endpoint in a reference section, continuous or otherwise? I believe @erunion is correct that these were simply used to render magic block markdown, and will be entirely replaced by the new engine.
> 
> This means anchors will continue to be generated for any H1-6 tag in the body of the document. Which is what you were trying to avoid for continuous reference sections, if I understand correctly @gratcliff?

@rafegoldberg  The intention is to remove anchors from our custom magic-block headers, specifically for continuous reference.

erunion:
@gratcliff @rafegoldberg Do we still need this, or was https://github.com/readmeio/api-explorer/pull/605 the better fix?

erunion:
Closing this out!

gratcliff:
Reopening this. @rafegoldberg I'm curious - is this a fine implementation, or do we want to somehow merge `opts` into `props`. I think there are arguments to both keeping things separate, and combining them!

packages/markdown/components/Heading/index.jsx
```diff
@@ -1,26 +1,35 @@
...
-    <a key={`heading-anchor-icon-${props.id}`} className="heading-anchor-icon fa fa-anchor" href={`#${props.id}`} />,
-  ]);
+  ];
+
+  if (splitReferenceDocs) {
+    children.push(
+      // eslint-disable-next-line jsx-a11y/anchor-has-content
+      <a key={`heading-anchor-icon-${props.id}`} className="heading-anchor-icon fa fa-anchor" href={`#${props.id}`} />
+    );
+  }
```
rafegoldberg, line 25: By this logic, wouldn't we end up hiding anchor icons on guides and custom pages too, since the default is false?

packages/markdown/index.js
```diff
@@ -202,12 +202,12 @@ export function react(text, opts = {}, components = {}) {
...
-        h3: Heading(3, count),
-        h4: Heading(4, count),
-        h5: Heading(5, count),
-        h6: Heading(6, count),
+        h1: Heading(1, count, opts),
+        h2: Heading(2, count, opts),
+        h3: Heading(3, count, opts),
+        h4: Heading(4, count, opts),
+        h5: Heading(5, count, opts),
+        h6: Heading(6, count, opts),
```
rafegoldberg, line 210: This passes in the full RDMD options object, yeah?

gratcliff:
> Read through in full; this looks good to me, but we've gotten enough grief about our heading anchors that I'm hesitant to approve this without pulling it down and fully testing it with my own 👀s. Coding with Emily on the free plan for a bit but will check out these changes this afternoon.

Works for me! I'd like to get a release out this afternoon with a couple things :)

packages/markdown/components/Heading/index.jsx
gratcliff: That's a good point, you may be right!

packages/markdown/index.js
gratcliff: Yeah.

packages/markdown/components/Heading/index.jsx
gratcliff: Fixed!

packages/markdown/components/Heading/index.jsx
```diff
@@ -1,26 +1,35 @@
 const React = require('react');
 const PropTypes = require('prop-types');
 
-function Heading({ tag, ...props }) {
+function Heading({ tag, opts, ...props }) {
   if (!props.children) return '';
 
+  const { splitReferenceDocs } = opts;
```
rafegoldberg, line 7:
```suggestion
  const { showAnchorIcons } = opts;
```

packages/markdown/components/Heading/index.jsx
```diff
@@ -29,6 +38,7 @@ Heading.propTypes = {
   children: PropTypes.array.isRequired,
   id: PropTypes.string.isRequired,
   level: PropTypes.number,
+  opts: PropTypes.object,
   tag: PropTypes.string.isRequired,
 };
 Heading.defaultProps = {
```
rafegoldberg, line 44:
```suggestion
Heading.defaultProps = {
  opts: { showAnchorIcons: true }
```

packages/api-explorer/src/block-types/Content.jsx
```diff
@@ -43,10 +43,10 @@ const Loop = ({ content, column, flags }) => {
 };
 
 const Content = props => {
-  const { body, isThreeColumn, useNewMarkdownEngine } = props;
+  const { body, isThreeColumn, useNewMarkdownEngine, splitReferenceDocs } = props;
 
   if (useNewMarkdownEngine) {
-    const content = markdown(body);
+    const content = markdown(body, { splitReferenceDocs });
```
rafegoldberg, line 49:
```suggestion
    const content = markdown(body, { showAnchorIcons: splitReferenceDocs });
```

packages/markdown/components/Heading/index.jsx
```diff
@@ -1,26 +1,35 @@
...
     <div key={`heading-anchor-${props.id}`} className="heading-anchor anchor waypoint" id={props.id} />,
     <div key={`heading-text-${props.id}`} className="heading-text">
       {props.children}
     </div>,
-    // eslint-disable-next-line jsx-a11y/anchor-has-content
-    <a key={`heading-anchor-icon-${props.id}`} className="heading-anchor-icon fa fa-anchor" href={`#${props.id}`} />,
-  ]);
+  ];
+
+  if (splitReferenceDocs === 'undefined' || splitReferenceDocs) {
```
rafegoldberg, line 20: I'm not sure this would work; wouldn't it need to be a `typeof` checK? In any case, I think we can simplify this logic [if we set the default for this prop to `true`]()!

```suggestion
  if (showAnchorIcons) {
```

PR #602, opened 2020-04-14
fix: issues preventing people from upgrading off our legacy explorer

erunion:
## 🧰 What's being changed?

This resolves a handful of issues that are preventing people from upgrading off our legacy API Explorer:

* [x] If we have a language that's technically unsupported by our code snippet generator (like JSON or HTTP), still make an attempt to capitalize the languages name by running it through our common `@readme/syntax-highlighter/uppercase` method. Without this, there might be a case where a language like JSON ends up in the Explorer's code tab (by way of the manual API editor example) as "json".
* [x] Stop nesting JSON response examples in a double `<div className="example example_json">`.
* [x] If a legacy response hasn't been fully configured yet, it will have a status code of `""` so we should ignore it so as to not render an empty tab in the `ResponseExample` component.
* [x] If a `requestBody` had a `title` or `description`, we'd retain that text and show some orphan content. (see https://github.com/readmeio/oas/pull/160)
* [x] If there was an OAS that was synced into our system that didn't have a `servers` object (something that could have been possible before we tightened down our OAS validation), we should still try to render the `PathUrl` for that as `@readme/oas-tooling` will set the base URL to `https://example.com` if no server base URL is present.

* [x] **🐛 I'm fixing a bug!**


packages/api-explorer/src/ResponseExample.jsx
```diff
@@ -246,7 +246,7 @@ class ResponseExample extends React.Component {
                         )}
 
                       {isJson && !example.multipleExamples ? (
-                        <div className="example example_json">{transformExampleIntoReactJson(example)}</div>
+                        transformExampleIntoReactJson(example)
```
rafegoldberg, line 249: ### 👏

packages/api-explorer/src/lib/upgrade-legacy-responses.js
```diff
@@ -13,6 +13,11 @@ module.exports = responses => {
   const codes = {};
 
   responses.forEach(response => {
+    // If there's no status set, then the legacy response wasn't full set up so we should ignore it.
```
rafegoldberg, line 16: these comments are great; minor sp. suggestion

```suggestion
    // If there's no status set, then the legacy response wasn't fully configured so we should ignore it.
```

PR #607, opened 2020-04-17
RDMD Feedback Fixes (4)

rafegoldberg:
## 🧰 What's being changed?

- [x] 🆕 Parse smart-sized magic image widths to HAST `hProperties`.
- [x] 🐛 Fix magic image null case (i.e. when no image src is set.)
- [x] 🐛 Normalize code tab default syntax names.
- [x] 🐛 Display single magic code blocks as tabs if a tab name is specified.
- [x] 🧪 Update various test snapshots.
- [x] 👊 Bump remark-stringify to v8.0.0 (a75cf77)
- [x] 👊 Bump remark-breaks to v1.0.5 (541478e)
- [x] 👊 Bump rehype-raw to v4.0.2 (b7583fa)


packages/markdown/components/CodeTabs/index.jsx
```diff
@@ -1,6 +1,8 @@
 const React = require('react');
 const PropTypes = require('prop-types');
 
+const langRemap = require('./lang-remap.json');
```
erunion, line 4: Instead of this, you can use `@readme/syntax-highlighter/uppercase` and do:

```
{meta || uppercase(lang)}
```

rafegoldberg: Ahhh, that’s great. Will update accordingly.

erunion: Maybe that uppercase method should be split out into separate library 🤔

packages/markdown/components/CodeTabs/index.jsx
```diff
@@ -21,9 +23,10 @@ const CodeTabs = props => {
       <div className="CodeTabs-toolbar">
         {children.map(({ props: pre }, i) => {
           const { meta, lang } = pre.children[0].props;
+          /* istanbul ignore next */
           return (
             <button key={i} onClick={e => handleClick(e, i)} type="button">
-              {meta || `(${lang || 'plaintext'})`}
+              {meta || `${!lang ? 'Text' : upper(lang) || lang}`}
```
erunion, line 29: `upper(lang)` will always return the `lang` if it's not a recognized language so I don't know if you need the `|| lang` here

rafegoldberg: Let's get rid of it...

```suggestion
              {meta || `${!lang ? 'Text' : upper(lang)}`}
```

PR #618, opened 2020-04-21
fix: legacy api upgrade issues

erunion:
## 🧰 What's being changed?

This resolves some issues that are preventing projects from upgrading off our legacy API Explorer:

* [ ] If a parameter is missing a `type`, instead of continuing to generate invalid JSON Schema we'll now add `type: string`. https://github.com/readmeio/oas/pull/167
* [ ] If a request body or component schema property is missing a `type`, add `type: string` like we now do on parameters. https://github.com/readmeio/oas/pull/168
  * https://productsup-component-docs.readme.io/reference?newApiExplorer=true#put_check-1

* [x] **🐛 I'm fixing a bug!**

### DIFF
--- a/packages/api-explorer/package-lock.json
+++ b/packages/api-explorer/package-lock.json
@@ -1333,9 +1333,9 @@
       "dev": true
     },
     "@readme/oas-tooling": {
-      "version": "3.3.1",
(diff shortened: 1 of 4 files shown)


erunion:
@kanadgupta Lerna loves making a mess sometimes. 😞

PR #629, opened 2020-04-25
RDMD magic block null cases.

### LINKED ISSUE #628: Better null case for magic tables.
If a table has no `data`, the AST parser should ignore it:

```
[block:parameters]
{
  "data": {},
  "cols": 3,
  "rows": 1
}
[/block]
```

rafegoldberg:
## 🧰 What's being changed?

I definitely thought I'd fixed all of these null cases, but I must have missed a few:

- [x] Skip empty magic table blocks (closes #628)
- [x] Skip empty magic code tab blocks
- [x] Skip empty magic callout blocks
- [x] Don't trim empty callout titles

## 🗳 Checklist

* 🐛 I'm fixing a bug

### DIFF
--- a/packages/markdown/processor/parse/magic-block-parser.js
+++ b/packages/markdown/processor/parse/magic-block-parser.js
@@ -126,6 +126,7 @@ function tokenize(eat, value) {
       };
       json.type = json.type in types ? types[json.type] : [json.icon || '👍', json.type];
       const [icon, theme] = json.type;
+      if (!(json.title || json.body)) return eat(match);
       return eat(match)(
         WrapPinnedBlocks(
           {
--- a/packages/markdown/processor/parse/magic-block-parser.js
+++ b/packages/markdown/processor/parse/magic-block-parser.js
@@ -51,7 +51,10 @@ function tokenize(eat, value) {
           },
         },
       }));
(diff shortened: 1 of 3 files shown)


packages/markdown/processor/parse/magic-block-parser.js
```diff
@@ -51,7 +51,10 @@ function tokenize(eat, value) {
           },
         },
       }));
-      if (children.length === 1 && children[0].name) return eat(match)(WrapPinnedBlocks(children[0], json));
+      if (children.length === 1) {
+        if (!children[0].value) return eat(match); // skip empty code tabs
```
rafegoldberg, line 55: for if, like, when a user drags in a code tab block to the end of a long doc, and then forgets about it but saves anyways, right? now that won't break shit.

packages/markdown/processor/parse/magic-block-parser.js
```diff
@@ -51,7 +51,10 @@ function tokenize(eat, value) {
...
+        if (children[0].name) return eat(match)(WrapPinnedBlocks(children[0], json));
+      }
```
rafegoldberg, line 57: These changes (and the following updates generally) are for if, like, when a user drags in a code tab block to the end of a long doc, and then forgets about it but saves anyways... Now dumb shit like that doesn't just break everything.

PR #630, opened 2020-04-25
RDMD image polish.

### LINKED ISSUE #621: Add lightbox functionality for images.
RDMD images no longer support lightbox functionality. We should bring this back for our users who rely on it!

### LINKED ISSUE #623: Inline images.
There's a bug in the base RDMD CSS that forces all images to display as block-level elements. Images should be displayed as inline blocks by default.

rafegoldberg:
## 🧰 What's being changed?

- [x] display images inline by default (closes #623)
- [x] add image lightbox component (close #621)

## 🗳 Checklist

- 🆕 I'm adding something new!
- 🐛 I'm fixing a bug!
- 📸 I've made some changes to the UI!

### DIFF
--- a/packages/markdown/components/Image/index.jsx
+++ b/packages/markdown/components/Image/index.jsx
@@ -12,7 +12,7 @@ class Image extends React.Component {
   }
 
   render() {
-    const { align, title, alt, width, height } = this.props;
+    const { align, title, alt, width, height, className = '' } = this.props;
     const extras = { align, width, height };
     return (
       <React.Fragment>
@@ -22,7 +22,7 @@ class Image extends React.Component {
           title={title}
           {...extras}
           loading="lazy"
-          onClick={() => this.setState({ lightbox: true })}
+          onClick={() => className !== 'emoji' && this.setState({ lightbox: true })}
         />
         <span
           className="lightbox"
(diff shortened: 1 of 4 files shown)


packages/markdown/components/Image/index.jsx
```diff
@@ -1,18 +1,52 @@
+/* eslint-disable jsx-a11y/click-events-have-key-events */
+/* eslint-disable jsx-a11y/no-noninteractive-element-interactions */
```
rafegoldberg, line 2: I think in the long run we should pull this kind of super UX-y type of behavior in to the UI. (Which is to say so I'm not too worried about the long-term accessibility concerns of this implementation.)

packages/markdown/components/Image/index.jsx
```diff
@@ -1,18 +1,52 @@
...
+            onClick={() => this.setState({ lightbox: false })}
+            onScrollCapture={() => this.setState({ lightbox: false })}
+            open={this.state.lightbox}
+            role="dialog"
+          >
+            <span className="lightbox-inner">
+              <img {...this.props} alt={alt} title="Click to close..." {...extras} loading="lazy" />
+            </span>
+          </span>
+        )}
```
rafegoldberg, line 39: This is a very simple, markup and CSS-driven lightbox. My goal here was mainly to keep the necessary Javascript as quick, simple, and lightweight (this requires zero additional deps!) as possible.

packages/markdown/components/Image/index.jsx
erunion: Yeah we should definitely have a `<Lightbox>` component in the UI that we could use here instead.

PR #648, opened 2020-04-28
fix: filter out nullish arrays from generated HAR objects

erunion:
## 🧰 What's being changed?

This resolves a funky case with HAR generation where if a user clicks add an entry to an `array<primitive>` parameter, we would immediately add `[null]` into their code sample.

Screen Shot 2020-04-28 at 12 27 38 PM

Why this was happening because if you supply `[undefined]` to `JSON.stringify`, it'll transform that into `[null]`, and we didn't have any work to filter out nullish items from arrays.

Screen Shot 2020-04-28 at 12 28 32 PM

## 🧪 Testing

Load up http://localhost:9966/?selected=swagger-files%2Ftypes.json, head to the "Arrays of different data types" operation and play around with adding and removing arrays.

## 🗳 Checklist
* [x] **🐛 I'm fixing a bug!**

### DIFF
--- a/packages/oas-to-har/__tests__/lib/remove-undefined-objects.test.js
+++ b/packages/oas-to-har/__tests__/lib/remove-undefined-objects.test.js
@@ -4,10 +4,19 @@ test('should remove empty objects with only undefined properties', () => {
   expect(removeUndefinedObjects({ a: { b: undefined, c: { d: undefined } } })).toBeUndefined();
 });
 
-test('should not throw on arrays of primitives', () => {
-  expect(removeUndefinedObjects([null])).toStrictEqual([null]);
+test('should remove empty arrays from within object', () => {
+  expect(removeUndefinedObjects({ a: { b: undefined, c: { d: undefined } }, d: [1234, undefined] })).toStrictEqual({
+    d: [1234],
+  });
 });
 
-test('should not throw for null', () => {
-  expect(removeUndefinedObjects({ a: null })).toStrictEqual({ a: null });
+test('should remove undefined and null values from arrays', () => {
+  expect(removeUndefinedObjects([undefined, undefined])).toBeUndefined();
+  expect(removeUndefinedObjects([null])).toBeUndefined();
+  expect(removeUndefinedObjects(['1234', null, undefined, { a: null, b: undefined }])).toStrictEqual([
+    '1234',
+    {
+      a: null,
+    },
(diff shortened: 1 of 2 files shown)


packages/oas-to-har/__tests__/lib/remove-undefined-objects.test.js
```diff
@@ -4,10 +4,21 @@ test('should remove empty objects with only undefined properties', () => {
   expect(removeUndefinedObjects({ a: { b: undefined, c: { d: undefined } } })).toBeUndefined();
 });
 
-test('should not throw on arrays of primitives', () => {
```
domharrington, line 7: Did you remove this test? Is it no longer required? Or it now returns undefined completely instead of `[null]`?

packages/oas-to-har/src/lib/remove-undefined-objects.js
```diff
@@ -5,26 +5,44 @@ function isEmptyObject(obj) {
...
-  // JSON.stringify removes undefined values
-  const withoutUndefined = JSON.parse(JSON.stringify(obj));
+  // JSON.stringify removes undefined values. Though `[undefined]` will be converted with this to
+  // `[null]`, we'll clean that up next.
+  let withoutUndefined = JSON.parse(JSON.stringify(obj));
 
-  // Then we recursively remove all empty objects
-  stripEmptyObjects(withoutUndefined);
+  // Then we recursively remove all empty objects and nullish arrays.
+  withoutUndefined = stripEmptyObjects(withoutUndefined);
```
domharrington, line 45: Is there a reason behind it now using a return value? I think I did it the previous way with mutable values because we weren't copying anything, we were messing with the original variable, so it seems like the truest representation of that.

packages/oas-to-har/__tests__/lib/remove-undefined-objects.test.js
erunion: Removed it because `[null]` should return `undefined` now.

packages/oas-to-har/src/lib/remove-undefined-objects.js
erunion: `obj.filter` needs to be assigned to something, and doing a re-assign to `obj` wasn't populating the reference.

Screen Shot 2020-04-28 at 12 49 24 PM

PR #660, opened 2020-04-30
Enable block-level components in magic tables.

rafegoldberg:
[Demo App](http://rgd2-group.readme-stage-pr-2700.readme.ninja/gh2700/docs/getting-started) |   [Issue](https://github.com/readmeio/api-explorer/issues/668) |   [ReadMe Integration](https://github.com/readmeio/readme/pull/2700)
:------:|:------:|:------:

## 🧰  Changes

Per the spec, only inline decorations are permitted within Markdown table cells. Our old magic table blocks did not heed this rule. This PR explores the necessary work and implications of implementing this functionality in the new engine. Technically, the level of effort is negligible — [](http://rgd2-group.readme-stage-pr-2700.readme.ninja/gh2700/docs/getting-started) I literally changed [one word](https://github.com/readmeio/api-explorer/compare/rdmd/table-block-contents?expand=1#diff-f0f0ac3f951c416b21abb269211108eaR172) to get it working!

That said, I'm still _extremely_ torn about whether or not we should allow it? I understand the value this level of customization offers our users. But I'm hesitant to move away from a "pure" markdown standard, with all the concomitant compatibility issues and long-term support overhead we now know come along with such a decision...

## 🧪  Testing

Both standard Markdown and ReadMe-flavored block components should render properly within magic table cells when the new compatibility mode is turned on:

- [**Broken** in production](https://rgd2-group.readme.io/gh2700/docs).
- [**Working** in the integration PR app](http://rgd2-group.readme-stage-pr-2700.readme.ninja/gh2700/docs/getting-started).


erunion:
I'm with you on this in that I can definitely see the benefit, but am not sure if it's best to stray too far from pure markdown.

That said, being able to add a bulleted list into a table is something I've always wished Markdown could do and am always frustrated to remember that it's not possible every time I try to add one into a table.

kanadgupta:
Here's a somewhat crazy idea... what if we had some project flag (that was disabled by default) that when enabled, allowed the user to do experimental anti-Markdown things like this? And we'd warn them prior to enabling the flag that if they deal with compatibility issues in the future when we switch to a new editor, it's... their own fault?

rafegoldberg:
@kanadgupta — I'd actually implemented something like this to allow for certain "bad behaviors", but I nixed it before the beta ended. I _do_ like it for this kind of backwards-compatibility feature, though! Maybe instead of "Experimental Mode" we can call it "Compatibility Mode" and use it as a place to cram in any/all functionality patches which we don't plan on continuing to support in the future?

erunion:
I like that, or treating it as the Markdown variant of JS `use strict` by allowing block-level components in the non-strict mode. Probably should enable strict by default though.

rafegoldberg:
So @erunion's mention of `use strict` got me thinking: what if, instead of having a God Mode admin flip this flag for a _whole_ project, an author could simply add some kind of `/* COMPAT MODE */` comment to the top of any doc that was misbehaving? (This could be interpreted and stripped by the engine itself! Would be nice to have everything self-contained, rather than having to set up some weird ReadMe flag and fork a bunch of logic!)

erunion:
I like that idea in concept, though it might be a bit more difficult for us have visibility into what pages are running in this mode since we'll need to do a regex query in Mongo against page bodies and potentially OAS file contents.

rafegoldberg:
(a brief update: after some further thought, a brief discussion with @amyshi188 a minute ago, and just generally following the continued saga of the RDMD rollout, I think something like this is very likely the best approach moving forwards with the various customers who have not been able to migrate fully to the new engine!)

trattner:
yes, seconded, PLEASE PLEASE HAVE THIS as it helps enormously with backwards compatibility!!!!! 

big sticking point with customers who don't want to manually change all their tables due to changes we made

packages/api-explorer/src/block-types/Content.jsx
```diff
@@ -124,7 +127,7 @@ Loop.defaultProps = {
 
 Content.propTypes = {
   body: PropTypes.string,
-  flags: PropTypes.shape({}),
+  flags: PropTypes.shape(),
```
erunion, line 130: This resolve a warning?

rafegoldberg: Yeah. I can't remember the specific wording, but essentially it seemed to be telling me that specifying a shape of an empty object actually sets an expectation that the passed value will have no keys.

PR #661, opened 2020-04-30
Add callout icon size and weight theme variables.

rafegoldberg:
- [x] 🆕 add callout `--icon-size` var
- [x] 🆕 add callout `--icon-weight` var
- [x] 🐷 reposition callout icons from top


rafegoldberg:
@kanadgupta 

<p>yeah, this is a  approved PR. Very low energy changes.</p>

PR #662, opened 2020-04-30
RDMD syntax to specify initial active code tab.

rafegoldberg:
Specify the initially visible tab by adding the `[open]` attribute after the code block's lang and title metadata. If multiple tabs are open, the last one will be visible. So in this example, the fourth tab would start opened on page load:

    ```js Tab 1
    First tab.
    ```
    ``` Tab 2 [open]
    Open tab, with title and no lang.
    ```
    ``` [open]
    Open tab, with no title or lang.
    ```
    ```php [open]
    Open tab, with no lang but no lang.
    ```
    ``` Last Tab
    Last tab.
    ```

This is compatible with magic blocks too, jic:

```
[block:code]
{
  "codes": [
    {
      "code": "hello world",
      "open": true
    }
  ]
}
[/block]
```

### DIFF
--- a/packages/markdown/components/Code/style.scss
+++ b/packages/markdown/components/Code/style.scss
@@ -136,13 +136,13 @@
       top: 50%;
       left: 50%;
       transform: translate(-50%, -50%) scale(.33);
+      color: var(--md-code-text, green) !important;
       opacity: 0 !important;
     }
 
     &_copied {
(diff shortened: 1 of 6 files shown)


rafegoldberg:
This work should be replicated in [the dedicated RDMD repo](https://github.com/readmeio/markdown/); closing here.

PR #664, opened 2020-04-30
Fix magic block callout icons alignment.

rafegoldberg:
|[Asana]()|
|:-:|

- [x] update pattern
- [x] add magic callout block test case

### Broken

image

### Fixed

image

## 🗳 Checklist

* 🐛 I'm fixing a bug

### DIFF
--- a/packages/markdown/components/Callout/index.jsx
+++ b/packages/markdown/components/Callout/index.jsx
@@ -4,6 +4,8 @@ const PropTypes = require('prop-types');
 const Callout = props => {
   let { children } = props;
   const { attributes, theme, title, icon } = props;
+  if (!(title || children)) return '';
+
   const content = title ? children.splice(1) : children;
   children = title ? children : '';
   return (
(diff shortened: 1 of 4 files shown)


packages/markdown/components/Callout/index.jsx
```diff
@@ -13,7 +15,7 @@ const Callout = props => {
         <span className="callout-icon">{icon}</span>
         {children}
       </h3>
-      {(content.length && content) || (!title ? children : '')}
+      {(content && content.length && content) || (!title && children)}
```
erunion, line 18: Think you have an extra content in here

rafegoldberg: Thought you were right, but check again and I believe this is what happens if:

1. I take out the first reference it flips out about not being able `to read "length" of undefined` when the callout has no body text
2. And taking out the second reference obviously doesn't work cause then it just prints the length of the variable, rather than the actual React tree!

PR #669, opened 2020-05-01
Fix RDMD backwards compatible anchors.

rafegoldberg:
- [x] 🐛 1-to-1 with our old anchor generator
- [x] 🚫 remove prefixed GH slug plugin
- [x] 🐷 fix deprecated anchors over-scrolling


rafegoldberg:
Actually one sec, I copied the `generateHeadId()` method from the old magic engine, but I believe the one we actually need to imitate is here:

packages/markdown/processor/plugin/section-anchor-id.js
```diff
@@ -1,17 +0,0 @@
-const flatMap = require('unist-util-flatmap');
```
domharrington, line 1: Make sure you remove this dependency if it's not being used anymore! (i can't remember if i added it or not)

rafegoldberg:
Actually, I copied the `generateHeadingId()` method from the old magic engine; I think the one we actually need to imitate is here:

https://github.com/readmeio/readme/blob/master/public/blocks/api-header/directive.js#L5-L12

PR #670, opened 2020-05-01
Fix RDMD lightbox bug.

rafegoldberg:
This CSS was causing repaint lag [on complex, long pages with a lot of images](https://readmeio.slack.com/archives/C04EF46CU/p1588272366221700) in Webkit browsers:

https://github.com/readmeio/api-explorer/blob/0577b73b4ef285a9b86bdde61bb9292bdd6a5768/packages/markdown/components/Image/style.scss#L83-L84

Removing these two lines fix this otherwise catastrophic seeming bug!


packages/markdown/components/Image/index.jsx
```diff
@@ -15,7 +15,7 @@ class Image extends React.Component {
     const { align, title, alt, width, height, className = '' } = this.props;
     const extras = { align, width, height };
     return (
-      <React.Fragment>
+      <div>
         <img
```
rafegoldberg, line 19: I think making such heavy use of the fragment component may also have been a contributing factor to this perf bottleneck. Definitely heavier than a native `<div/>`.

rafegoldberg:
one of these days I'll remember to `npm run test`s before I request review but that day a'int today!

PR #679, opened 2020-05-05
feat: nested ast processing

gratcliff:
## 🧰 What's being changed?
1. Adding support for deep-table AST processing. Functionally bringing values to the top child level within a `table` node.
2. Adding a new export, `astToPlainText` that returns plain text from an AST node.

## 🧪 Testing
This should not effect ReadMe's Hubs, no differences will appear there. Instead, you can test the lib's export methods via passing a fresh AST node and observing the result.

* [X ] **🆕 I'm adding something new!**
* [ ] **🐛 I'm fixing a bug!**


packages/markdown/components/Heading/index.jsx
```diff
@@ -1,14 +1,5 @@
 const React = require('react');
 const PropTypes = require('prop-types');
-const kebab = require('lodash/kebabCase');
```
erunion, line 3: Do we still need lodash?

gratcliff: Yup, it's being used in section hash hrefs!

erunion: Ah, missed that you moved it.

gratcliff: In one of our plugin transformers.

packages/markdown/processor/compile/index.js
```diff
@@ -4,3 +4,4 @@ export { default as rdmeEmbedCompiler } from './embed';
 export { default as rdmeVarCompiler } from './var';
 export { default as rdmeCalloutCompiler } from './callout';
 export { default as rdmePinCompiler } from './pin';
+export { default as plainTextCompiler } from './plain-text';
```
rafegoldberg, line 7: I think of this folder as being exclusively for methods that modify the actual Unified `Compiler` instance, like so:

https://github.com/readmeio/api-explorer/blob/66eeca17c49cf30d27cbf1277ea7118bdac9a91a/packages/markdown/processor/compile/callout.js#L3

Can we move this `plain-text.js` script to the `plugin/` directory? Should work just as well from there!

packages/markdown/processor/plugin/section-anchor-id.js
```diff
@@ -0,0 +1,16 @@
+const flatMap = require('unist-util-flatmap');
+const kebabCase = require('lodash/kebabCase');
+
+// Adds an empty <div id='section-slug'> next to all headings.
+// This is for compatibility with how we used to do slugs
+function transformer(ast) {
+  return flatMap(ast, node => {
+    if (node.type === 'element' && node.properties && node.properties.id) {
+      const id = `section-${kebabCase(node.properties.id)}`;
+      return [node, { type: 'element', tagName: 'div', properties: { id } }];
```
rafegoldberg, line 10: One of the issues with Dom's initial take on this is that when you used the backwards-compatible hash link, the page would scroll _past_ the actual header Ideally we'd move this in to the node's `children` array. Not sure, but something like this...

```js
  // etc.
  if (/* blah && blah */) {
    node.children.push({ type: 'element', tagName: 'div', properties: { id } });
  }
  return [node];
}
```

...might work?

packages/markdown/processor/plugin/table-flattening.js
```diff
@@ -0,0 +1,39 @@
+const flatMap = require('unist-util-flatmap');
+
+// Flattens table values and adds them as a seperate, easily-accessible key within children
```
rafegoldberg, line 3: ### 💖
be still, my heart

packages/markdown/processor/plugin/section-anchor-id.js
gratcliff: Ahh that's interesting. Let me test it out!

packages/markdown/processor/compile/index.js
gratcliff: Happy to move it.

packages/markdown/processor/plugin/section-anchor-id.js
gratcliff: Screen Shot 2020-05-05 at 5 44 30 PM

gratcliff: @rafegoldberg updated.

gratcliff:
@erunion mind another once over and tagging a release if you're content?

erunion:
```
Successfully published:
 - @readme/api-explorer@6.8.0
 - @readme/markdown@6.8.0
```

Make sure this doesn't go out before @rafegoldberg's 6.7.0 work in https://github.com/readmeio/readme/pull/2551

gratcliff:
> ```
> Successfully published:
>  - @readme/api-explorer@6.8.0
>  - @readme/markdown@6.8.0
> ```
> 
> Make sure this doesn't go out before @rafegoldberg's 6.7.0 work in [readmeio/readme#2551](https://github.com/readmeio/readme/pull/2551)

Thanks! We're gonna jump a minor version in that PR.

PR #683, opened 2020-05-06
feature: prompt to log into docs for hubs with logs

kanadgupta:
## 🧰 What's being changed?

For hubs/explorers logs enabled, we'll encourage logged-out users to log into the hubs to see their logs:

image

## 🧪 Testing

I'll open up a PR in ReadMe, but to test in this repo:
1. Check out this branch
2. In in `Demo.jsx`, uncomment this line:
https://github.com/readmeio/api-explorer/blob/9af939466394d5ebeb62bea74eabbd89f4856e5a/example/src/Demo.jsx#L114
and comment out these lines:
https://github.com/readmeio/api-explorer/blob/9af939466394d5ebeb62bea74eabbd89f4856e5a/example/src/Demo.jsx#L124-L129
3. Run `npm start` from the root of this repo.

* [ ] **🐛 I'm fixing a bug!**


packages/api-explorer/src/Doc.jsx
```diff
@@ -418,6 +419,7 @@ Doc.propTypes = {
   ),
   language: PropTypes.string.isRequired,
   lazy: PropTypes.bool,
+  loginUrl: PropTypes.string,
```
erunion, line 422: Can you add a `.isRequired` to this or set `https://dash.readme.com/login` as the default?

kanadgupta: Hmm yeah I was waffling on how to handle this since technically it won't break anything if it's omitted. I'll add a default in the Log component.

kanadgupta: Updated: https://github.com/readmeio/api-explorer/pull/683/commits/b3fe21d11acf42ffac17950bc9428dd4323b2946

packages/api-logs/index.jsx
```diff
@@ -222,8 +222,36 @@ class Logs extends React.Component {
...
+                <path d="M3.29102 11.5893C3.29102 11.2069 3.59388 10.897 3.96749 10.897H14.791C15.1646 10.897 15.4675 11.2069 15.4675 11.5893C15.4675 11.9716 15.1646 12.2816 14.791 12.2816H3.96749C3.59388 12.2816 3.29102 11.9716 3.29102 11.5893Z" />
+                <path d="M3.29102 14.3586C3.29102 13.9762 3.59388 13.6663 3.96749 13.6663H9.37925C9.75286 13.6663 10.0557 13.9762 10.0557 14.3586C10.0557 14.7409 9.75286 15.0509 9.37925 15.0509H3.96749C3.59388 15.0509 3.29102 14.7409 3.29102 14.3586Z" />
+                <path
+                  clipRule="evenodd"
+                  d="M21.9895 1.70441H2.18145V16.4377H21.9895V1.70441ZM2.18145 0.0710449C1.3 0.0710449 0.585449 0.802327 0.585449 1.70441V16.4377C0.585449 17.3398 1.3 18.071 2.18145 18.071H21.9895C22.8709 18.071 23.5854 17.3398 23.5854 16.4377V1.70441C23.5854 0.802327 22.8709 0.0710449 21.9895 0.0710449H2.18145Z"
+                  fillRule="evenodd"
+                />
+              </svg>
+              <p>Log in to view your API logs for this endpoint</p>
+              <a className="logs-login-button" href={loginUrl}>
```
erunion, line 248:
```suggestion
              <a className="logs-login-button" href={loginUrl} rel="noreferrer noopener">
```

packages/api-logs/index.jsx
```diff
@@ -222,8 +222,36 @@ class Logs extends React.Component {
   }
 
   render() {
-    const { query, baseUrl, group } = this.props;
-    if (!group) return null;
+    const { query, baseUrl, group, loginUrl } = this.props;
+    if (!group)
+      return (
+        <div className="logs">
+          <div>
```
erunion, line 229: Should this have the `log-header` class?

kanadgupta: Nope! That class only [adds styles](https://github.com/readmeio/api-explorer/blob/2cd469a052a63fbc2c67386c3a49214e2573ea58/packages/api-logs/main.css#L25-L30) that make sense if there are logs present; e.g. a bottom border, spacing for this group selector dropdown (see below), etc.
image

packages/api-logs/index.jsx
kanadgupta: Do we need this since it's opening it in the same tab? I figured it'd be clunky to add `target="_blank"`, etc. since logging in would redirect back to the documentation.

erunion: Good point! It's only needed if we're opening in a new tab.

erunion: Also tested this on https://mathiasbynens.github.io/rel-noopener/ just to make sure, and it checks out. We don't need it here.

PR #684, opened 2020-05-07
fix(markdown): IE11 support

### LINKED ISSUE #680: IE syntax error.
|[Ticket](https://app.asana.com/0/681252538274980/1174342354720957)|
|-|

IE11 and below seem to encounter a syntax error when using Unified. Apparently you have to transpile it yourself if you want to use it in the browser per unifiedjs/unified#34

rafegoldberg:
## 🧰 What's being changed?

- [x] update webpack configs to transpile unified (resolves #680)
- [x] add style fallbacks for codeblocks on IE11
- [x] replace emojis with escaped unicode references for IE11
- [x] fix IE native method compatibility in the CodeTabs component

## 🗳 Checklist

* 🐛 I'm fixing a **huge** bug!


packages/markdown/components/CodeTabs/index.jsx
```diff
@@ -7,8 +7,10 @@ const CodeTabs = props => {
   const { attributes, children } = props;
 
   function handleClick({ target }, index) {
-    const $wrap = target.closest('.CodeTabs');
-    $wrap.querySelectorAll('.CodeTabs_active').forEach(el => el.classList.remove('CodeTabs_active'));
+    console.log(target);
```
erunion, line 10:
```suggestion
```

packages/markdown/processor/parse/flavored/callout.js
```diff
@@ -12,15 +12,19 @@ function tokenizer(eat, value) {
...
+    '\u26A0\uFE0F': 'warn',
+    '\uD83D\uDEA7': 'warn',
+    '\uD83D\uDC4D': 'okay',
+    '\u2705': 'okay',
+    '\u2757': 'error',
+    '\u2757\uFE0F': 'error',
+    '\uD83D\uDED1': 'error',
+    '\u2049\uFE0F': 'error',
+    '\u203C\uFE0F': 'error',
+    // eslint-disable-next-line prettier/prettier
```
erunion, line 25: What's this disabling?

packages/markdown/__tests__/__snapshots__/components.test.js.snap
```diff
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Components Callout 1`] = `"<blockquote class=\\"callout callout_error\\" theme=\\"❗️\\"><h3 class=\\"callout-heading false\\"><span class=\\"callout-icon\\">❗️</span><p>Error Callout</p></h3><p>Lorem ipsum dolor.</p></blockquote>"`;
+exports[`Components Callout 1`] = `"<blockquote class=\\"callout callout_error\\" theme=\\"❗\\"><h3 class=\\"callout-heading false\\"><span class=\\"callout-icon\\">❗</span><p>️ Error Callout</p></h3><p>Lorem ipsum dolor.</p></blockquote>"`;
```
erunion, line 3: Should there be a space at the top of the `<p>`, or between the emoji and paragraph?

packages/markdown/processor/parse/flavored/callout.js
rafegoldberg: ohhh shit, so I meant to leave a comment asking you this exact thing.

rafegoldberg: so whenever I run `npm test -- -u` on the repo, that line auto-prettifies itself _back_ to the ℹ️ emoji! I'm not quite sure what prettier rule it is that's catching this; maybe it's just some funky VSC setting on my machine? But my fear is that even with this disable comment, anyone who updates the snapshots for the repo will break the next release build in IE again without really noticing!

erunion: Huh, yeah. Weird.

Screen Shot 2020-05-07 at 9 03 05 AM

packages/markdown/components/CodeTabs/index.jsx
rafegoldberg: Oh word, thanks. And all of the emoji stuff will still work the same for the end user, but we could probably split the breaking syntax stuff out in to a separate script and fork the logic in IE, but between webpack config and IE debug, not sure it's worth the pain!

packages/markdown/processor/parse/flavored/callout.js
erunion: Leaving the Prettier exclusion in should be fine for future builds, but yeah that is concerning because:

```
> '\u2139\uFE0F' === 'ℹ️'
true
```

packages/markdown/__tests__/__snapshots__/components.test.js.snap
rafegoldberg: It's because the unicode for the ❗️emoji seems to include a trailing space to regex, hence the extra preceding space. But other than IE, this doesn't effect the actual render.

PR #685, opened 2020-05-07
chore: cleaning up some logic blocks in oas-to-har

erunion:
## 🧰 What's being changed?

Minor cleanup of some logic blocks within our `oas-to-har` library.


erunion:
Renaming this branch because I'm going to be doing some more work than just cleanup.

PR #686, opened 2020-05-07
fix/table-parsing

gratcliff:
## 🧰 What's being changed?

Fixing issue where nested table parsing failed from null hAST values. We're seeing some extremely large custom markdown tables with missing values. When they are serialized to hAST, and subsequently parsed here, we're not covering the case that they don't exist.

Now we are!

* [X ] **🐛 I'm fixing a bug!**

### DIFF
--- a/packages/markdown/processor/plugin/table-flattening.js
+++ b/packages/markdown/processor/plugin/table-flattening.js
@@ -9,10 +9,11 @@ function transformer(ast) {
       // This is necessary to pullout all the relevant strings
 
       // Parse Header Values
-      const headerChildren = header.children[0].children;
+      const headerChildren = header.children && header.children.length ? header.children[0].children : [];
       const headerValue = headerChildren.map(hc => (hc.children.length && hc.children[0].value) || '').join(' ');
       // Parse Body Values
-      const bodyChildren = body.children.map(bc => bc.children).reduce((a, b) => a.concat(b), []);
+      const bodyChildren =
+        (body.children && body.children.map(bc => bc && bc.children).reduce((a, b) => a.concat(b), [])) || [];
       const bodyValue = bodyChildren.map(bc => (bc.children.length && bc.children[0].value) || '').join(' ');
 
       return [
@@ -37,3 +38,4 @@ function transformer(ast) {
 }
 
 module.exports = () => transformer;
+module.exports.tableFlattening = transformer;
--- a/packages/markdown/__tests__/index.test.js
+++ b/packages/markdown/__tests__/index.test.js
@@ -3,6 +3,7 @@ const React = require('react');
 const BaseUrlContext = require('../contexts/BaseUrl');
 
 const markdown = require('../index');
+const { tableFlattening } = require('../processor/plugin/table-flattening');
 const settings = require('../options.json');
 
 test('image', () => {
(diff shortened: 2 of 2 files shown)


packages/markdown/processor/plugin/table-flattening.js
```diff
@@ -9,10 +9,11 @@ function transformer(ast) {
       // This is necessary to pullout all the relevant strings
 
       // Parse Header Values
-      const headerChildren = header.children[0].children;
+      const headerChildren = header.children && header.children.length ? header.children[0].children : [];
```
rafegoldberg, line 12: Are we allowed to use the null coalescing operator yet? Syntax is still new to me, but would this work?

```suggestion
      const headerChildren = header?.children?.[0].children || [];
```

packages/markdown/processor/plugin/table-flattening.js
```diff
@@ -37,3 +38,4 @@ function transformer(ast) {
 }
 
 module.exports = () => transformer;
+module.exports.tableFlattening = transformer;
```
rafegoldberg, line 41: Why the addition?

gratcliff: Testing!

packages/markdown/processor/plugin/table-flattening.js
gratcliff: @erunion do you know?

erunion: Optional chaining was turned on by default in `@babel/preset-env` 7.8, and we're on 7.9 so you should be good to use it! Would recommend looking at the compiled dist and making sure it's translated okay though.

https://github.com/babel/babel/pull/10811

gratcliff: I'll give it a shot!

erunion: Assuming it's good, once ESLint or [eslint-plugin-unicorn](https://www.npmjs.com/package/eslint-plugin-unicorn) get a rule for it, we can probably start to enforce and prefer optional chaining

gratcliff: Yeah we definitely don't support it in our linting at this moment. Might want to include it [here](https://github.com/readmeio/eslint-config/blob/master/index.js) going forward?

gratcliff: Ah nevermind, I see what you're saying. 
> Once it's included. 
Let's hold off on moving this until we can lint enforce it.

rafegoldberg: good by me 👍

gratcliff:
@erunion @rafegoldberg we good on this?

packages/markdown/processor/plugin/table-flattening.js
erunion: Yeah I meant you should be good to use it now, just that once it gets added into ESLint we'll start enforcing it in our apps that use Babel.

rafegoldberg:
Yeah, I was just playing around! All good by me.

packages/markdown/processor/plugin/table-flattening.js
gratcliff: I got ya. Let's just do it all in one sweep!

erunion:
Yeah lgtm

PR #700, opened 2020-05-12
Better code block dark theme in response schema sidebar.

rafegoldberg:
## What's being changed?

- [x] 🐷 wrap sidebar response code samples with a `.markdown-body` class parent
- [x] 🧬 update response schema tests

The before & after:

 

## Testing

Will open an integration PR in ReadMe and post a link to the demo app.

### DIFF
--- a/packages/api-explorer/__tests__/ResponseSchema.test.jsx
+++ b/packages/api-explorer/__tests__/ResponseSchema.test.jsx
@@ -37,11 +37,10 @@ test.each([[true], [false]])(
     let text;
 
     if (useNewMarkdownEngine) {
-      text = responseSchema.find('div.desc').first().find('div.desc').find('p').first().text();
+      text = responseSchema.find('.markdown-body > .pin > *').first().text();
     } else {
       text = responseSchema.find('div.desc').first().text();
     }
-
     expect(text).toBe(props.operation.responses['200'].description);
   }
 );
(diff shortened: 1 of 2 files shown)


erunion:
Is that still using the tomorrow-night theme that we use in `ResponseBody`?

rafegoldberg:
@erunion in short? no, they use an "inverted" style. This is because of the somewhat hacky way the RDMD engine is handling pinned blocks at the moment: we always render the full doc (pinned and non-pinned blocks alike), [wrapping sidebar blocks](https://github.com/readmeio/api-explorer/blob/master/packages/markdown/processor/parse/magic-block-parser.js#L4-L14) in a `<div.pin/>` as we go. Then we [use CSS to conditionally](https://github.com/readmeio/api-explorer/blob/06aa341b629680b255e1469e427c76fe08e05bc3/packages/api-explorer/style/markdown-overrides.scss#L21-L28) display these blocks depending on a few upper-level selector conditions.

This makes it somewhat tricky to determine the right theme in advance and do things the same way it was set up before. So yeah, we def. _should_ refactor this to use a real dark theme; I can imagine the steps look something like:

- [ ] fork the `rdmd.react` processor export so it can return:
  - the entire content tree
  - only the main content tree
  - pinned sidebar content only
- [ ] update various RDMD React integrations
  - across the Explorer
  - in ReadMe

But it'll take some more thought and work to integrate the prop passing correctly across all of the RDMD code blocks. This is a cheap solution _and_ it'll hold quite well for now (at least right up until one of our users wants to customize their syntax highlighting theme...)

PR #705, opened 2020-05-22
feat: fixing some ssr issues with the markdown renderer 

domharrington:
## 🧰 What's being changed?

feat: fixing some ssr issues with the markdown renderer
- always render the copy code button, even on the server
Fixes this issue `Warning: Expected server HTML to contain a matching <button> in <code>.`

- Added `suppressHydrationWarning={true}` to the `<code>` element
The rendered content is always going to be different due to one having
syntax highlighting and one not

## 🧪 Testing

`npm test` - i've actually removed a code path here that wouldn't have previously been hit in our test suite. Dont think any other tests need to be written for this.

* [x] **🐛 I'm fixing a bug!**


packages/markdown/processor/parse/flavored/callout.js
```diff
@@ -22,9 +22,7 @@ function tokenizer(eat, value) {
     '\uD83D\uDED1': 'error',
     '\u2049\uFE0F': 'error',
     '\u203C\uFE0F': 'error',
-    // NOTE: prettier is desperate to convert this in to an emoji.
```
erunion, line 25: Can you leave this comment in?

rafegoldberg: Assuming this updated `prettier-ignore` statement works I’m comfortable deleting my scary comment!

erunion: Yeah, I just want to leave the scary comment so we know why the prettier ignore is there.

PR #714, opened 2020-05-28
Strip preceding dashes from anchor IDs.

### LINKED ISSUE #713: Don't start slugs with a `-`!
If you start slugs with emojis, they get stripped from the slugs! However, the slug looks like `#-something`. Would it be possible to make it so that slugs never start with `-`?

Here's the header:
image

And here's the slug it generates:

image

rafegoldberg:
Polish some more rough edges around section anchor generation. Specifically:

- [x] strip prefixed dashes from section anchors; resolves #713 (ac7f762)
- [x] use compat anchor IDs as fallback when the GitHub slugger returns an empty string (b2fdac4)

### DIFF
--- a/packages/markdown/processor/plugin/section-anchor-id.js
+++ b/packages/markdown/processor/plugin/section-anchor-id.js
@@ -20,16 +20,29 @@ const getTexts = node => {
 function transformer(ast) {
   return flatMap(ast, node => {
     if (matchTag.test(node.tagName)) {
-      if (node?.properties?.id) node.properties.id = node.properties.id.replace(/^(-+(?=\w))/, '');
+      // Parse the node texts to construct
+      // a backwards-compatible anchor ID.
       const text = getTexts(node);
       const id = `section-${kebabCase(text)}`;
-      const element = {
+
+      if (node?.properties?.id) {
+        // Strip dash prefixes in GitHub slugger IDs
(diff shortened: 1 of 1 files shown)


packages/markdown/processor/plugin/section-anchor-id.js
```diff
@@ -20,14 +20,29 @@ const getTexts = node => {
...
-      const id = `section-${kebabCase(getTexts(node))}`;
-      const element = {
+      // Parse the node texts to construct
+      // a backwards-compatible anchor ID.
+      const text = getTexts(node);
+      const id = `section-${kebabCase(text)}`;
+
+      if (node?.properties?.id) {
+        // Strip dash prefixes in GitHub slugger IDs
+        node.properties.id = node.properties.id.replace(/^(-+(?=\w))/, '');
```
erunion, line 30: Should this also update `id` to match this? Doesn't seem right that `node.properties.id` wouldn't match the `properties.id` that's set on the anchor on line 39.

rafegoldberg: Our new [slug generator](https://github.com/Flet/github-slugger) isn't an exact one-to-one with our old anchors, so we have to provide for backwards compatibility. We do that here by generating an additional anchor using our old, lodash-based strategy and injecting that in to the header HAST. In short these are intentionally different!

PR #727, opened 2020-06-02
Rearchitecting how the syntax highlighter library is unit tested

erunion:
## 🧰 What's being changed?

This completely rearchitects how `@readme/syntax-highlighter` is unit tested by creating a new drop-in test system that makes adding support for new languages TDD-driven. Now when adding a new language, you can create a new fixture, with a code sample, in `__tests__/__fixtures__/newLanguage` and the unit testing architecture will automatically detect it and run test assertions against it.

For details how it works, check the new contributing section in the README.

I've also added support for Objective-C++ and XHTML, as well as the following new language aliases:

* `aspx` for ASP.NET (and renamed ASP to ASP.NET in the process since that's its proper name)
* `m` for Objective-C
* `ps1` for PowerShell
* `styl` for Stylus

Finally, I've bulked up the README a bit to document all of these languages, and mode aliases, that we support. 🚀 

* [ ] **🐛 I'm fixing a bug!**


packages/syntax-highlighter/__tests__/__fixtures__/dockerfile/sample.dockerfile
```diff
@@ -0,0 +1 @@
+FROM alping:3.4
```
gratcliff, line 1: Should this be `alpine`?

erunion: 🥇

packages/syntax-highlighter/__tests__/__fixtures__/objc/sample.m
```diff
@@ -0,0 +1,14 @@
...
+ Build on OS X:
+ clang -framework Foundation -fobjc-arc objc.m -o objc
+
+ Build on Linux with GNUstep:
+ clang \`gnustep-config --objc-flags\` \`gnustep-config --base-libs\` -fobjc-nonfragile-abi -fobjc-arc objc.m -o objc
+ */
+
+#import <Foundation/Foundation.h>
+
+int main(void)
```
gratcliff, line 11: Do you need to pass in a `void` arg here? I think if you're trying to say it has no return value, the syntax is a bit different.

packages/syntax-highlighter/__tests__/index.test.js
```diff
@@ -1,5 +1,12 @@
 const { mount, shallow } = require('enzyme');
+const path = require('path');
+const glob = require('glob');
+const fs = require('fs').promises;
+
 const syntaxHighlighter = require('..');
+const uppercase = require('../uppercase');
+
+const fixtures = glob.sync(path.join(__dirname, '/__fixtures__/*'));
```
gratcliff, line 9: Lovely!!

packages/syntax-highlighter/__tests__/__fixtures__/objc/sample.m
erunion: Pulled this example from https://github.com/leachim6/hello-world/blob/master/o/objective-c.m, which reminds me I need to pull in their license.

Issue #726, opened 2020-06-02
Add support for missing languages


erunion:
We used to support syntax highlighting on the following languages but no longer:

- CoffeeScript
- Clojure
- Cypher
- D
- Erlang
- Groovy
- Handlebars
- Haxe
- HTTP
- Jinja
- Julia
- LESS
- Lua
- Perl
- Rust
- SASS
- Smarty
- Text
- TOMP
- Twig
- YAML

PR #728, opened 2020-06-03
feat: add support for syntax highlighting more languages 

erunion:
## 🧰 What's being changed?

Adds support to syntax highlight the following languages:

- [x] CoffeeScript
- [x] Clojure
- [x] D
- [x] Erlang
- [x] Groovy
- [x] Handlebars
- [x] HTTP
- [x] Julia
- [x] LESS
- [x] Perl
- [x] Rust
- [x] SASS
- [x] YAML

Languages from https://github.com/readmeio/api-explorer/issues/726 that I'm opting to not implement right now because their popularity is limited:

- Cypher
- Haxe
- Jinja
- Lua
- Smarty
- TOML
- Twig

I've also slightly refactored the `Code` component in `@readme/markdown` to no longer maintain a separate syntax highlighting language mapping object, and instead to pull that data from the syntax highlighter itself via a new `canonical` library within it. This library takes in a language name and returns a canonical version of it. For example, supplying `js` will return `javascript`, `cpp` will return `cplusplus`, `sass` to `css`, etc. This should make adding support for new languages less tedious.

Resolves https://github.com/readmeio/api-explorer/issues/726.

* [ ] **🐛 I'm fixing a bug!**


packages/markdown/processor/parse/flavored/callout.js
```diff
@@ -24,8 +24,7 @@ function tokenizer(eat, value) {
     '\u203C\uFE0F': 'error',
     // NOTE: prettier is desperate to convert this in to an emoji.
     //       PLEASE DON'T COMMIT THIS OR YOU'LL BREAK README IN IE!
-    // eslint-disable-next-line prettier/prettier
-    '\u2139\uFE0F': 'info',
+    '\u2139\uFE0F': 'info', // prettier-ignore
```
erunion, line 27: Running `npm run prettier` flipped this back to the ℹ️ emoji, so slapping `prettier-ignore` on this should finally do the trick.

😬

erunion:
Once this is out we can replace the [Language Support](https://docs.readme.com/rdmd/docs/code-blocks#language-support) docs on the RDMD site directly with the language support docs in `@readme/syntax-highlighter`. ✨

packages/syntax-highlighter/__tests__/__fixtures__/powershell/index.js
```diff
@@ -1,9 +1,9 @@
 module.exports = {
   language: 'PowerShell',
   mode: {
-    primary: 'ps1',
+    primary: 'powershell',
     aliases: {
-      powershell: 'PowerShell',
+      ps1: 'PowerShell',
```
erunion, line 6: https://www.youtube.com/watch?v=7Ret564LSLw

packages/syntax-highlighter/canonical.js
```diff
@@ -0,0 +1,60 @@
+const modeAliases = {
```
erunion, line 1: It might be kind of tedious to maintain this list along with `uppercase.js` instead of just a single `supportedLanguages` file, but also I think the object management, and cycles, that'll need to be done to make a single datastore work for both of these usecases (canonical and uppercase) outweighs just loading another 60 line JS file.

packages/markdown/processor/parse/flavored/callout.js
rafegoldberg: Oy vey. I need an `:eye-roll:` emoji to deal with this prettier stuff.

packages/syntax-highlighter/__tests__/__fixtures__/powershell/index.js
rafegoldberg: epic

packages/syntax-highlighter/canonical.js
rafegoldberg: Yeah the manual maintenance is a bit of a pain, but I agree: worth it compared to the alternative!

PR #765, opened 2020-06-04
Downgrade markdown-magic deps and fix tests.

rafegoldberg:
## 🧰 What's being changed?

This PR reverts the dep updates made to the `markdown-magic` sub-package in #759 & #760, and updates some test snapshots to account for class additions to checkbox task list items.

My bigger question is whether Dependbot would honor the `package.json` if we were to peg these deps to a specific version? (My guess is it would work, but not quite sure how that would look with Lerna!)
## 🧪 Testing

Test should run successfully.

## 🗳 Checklist
- I'm fixing some tests


erunion:
Yeah if you peg the version in `package.json` to a specific, non-semver version, Dependabot won't try to update it.

PR #807, opened 2020-06-26
fix: issue where application/x-www-form-urlencoded payloads weren't being handled properly

domharrington:
## 🧰 What's being changed?

This resolves a couple bugs with how we're generating HAR definitions:

* [x] For `application/x-www-form-urlencoded` operations, we were storing their payloads inside of `postData.text` instead of `postData.params`.
  * This required an update to [fetch-har](https://npm.im/fetch-har) to support `postData.params`. See https://github.com/readmeio/fetch-har/pull/67 for details.
* [x] We were always sending `postData` regardless if there was actually any data for it. For simple GET requests that just have `queryString` populated, the `postData` object is no longer part of the created HAR.
  * This required an update to [fetch-har](https://npm.im/fetch-har) to support `postData` no longer being an expectation. See https://github.com/readmeio/fetch-har/pull/67 for details.
* [x] The HAR definitions that `oas-to-har` was creating didn't actually validate outside of [httpsnippet](https://github.com/Kong/httpsnippet). This was because that library adds some defaults (like `httpVersion`) if they're missing. Since `oas-to-har` should support the HAR spec and not HTTP Snippets handling of it, I've made the decision to:
  * Add those missing defaults to generated definitions.
  * Update the unit test suite to assert that we're always generating valid HAR definitions when running our other assertions.

## 🧪 Testing

You can see this broken in action here: http://bin.readme.com/s/5efbb6ac35bcd40024d60488

Screen Shot 2020-06-30 at 3 04 39 PM

Load up the local server and the `formData` example to see how it's handled how: http://localhost:9966/?selected=swagger-files%2Fformdata.json

Screen Shot 2020-06-30 at 3 04 48 PM

* [x] **🐛 I'm fixing a bug!**

---
I added two test cases: one for json and one for formData. The json
one passes fine! This code in oas-to-har is working as expected:
https://github.com/readmeio/api-explorer/blob/59a5d060997321936ae356f2ba3f36e34c81c520/packages/oas-to-har/src/index.js#L198

I narrowed down the issue to here:
https://github.com/Kong/httpsnippet/blob/633b825364ab050f02556cfa837148a1e136fd69/src/index.js#L126-L128
httpsnippet is setting it from the value we set back to ""

I'm not sure if this is something that's changed recently in the way
we're generating hars? This part of httpsnippet hasn't been updated
in 5 years. Gunna defer to Jon for this.

### DIFF
--- a/packages/oas-to-har/__tests__/index.test.js
+++ b/packages/oas-to-har/__tests__/index.test.js
@@ -1,4 +1,4 @@
-const querystring = require('querystring');
+const validate = require('har-validator');
 const extensions = require('@readme/oas-extensions');
 const Oas = require('@readme/oas-tooling');
 
@@ -7,16 +7,40 @@ const commonParameters = require('./__fixtures__/common-parameters');
 
 const oas = new Oas();
 
-test('should output a har format', () => {
-  expect(oasToHar(oas)).toStrictEqual({
+expect.extend({
+  toBeAValidHAR(har) {
+    return validate
+      .request(har.log.entries[0].request)
+      .then(() => {
+        return {
+          message: () => `expected supplied HAR not to be valid`,
+          pass: true,
+        };
+      })
+      .catch(err => {
+        return {
+          message: () => `expected supplied HAR to be valid\n\nError: ${this.utils.printReceived(err.errors)}`,
+          pass: false,
+        };
+      });
+  },
+});
+
+test('should output a har format', async () => {
+  const har = oasToHar(oas);
+
+  await expect(har).toBeAValidHAR();
+  expect(har).toStrictEqual({
     log: {
       entries: [
         {
           request: {
+            bodySize: 0,
             cookies: [],
             headers: [],
+            headersSize: 0,
+            httpVersion: 'HTTP/1.1',
             method: '',
-            postData: {},
             queryString: [],
             url: 'https://example.com',
           },
@@ -26,8 +50,11 @@ test('should output a har format', () => {
   });
 });
 
-test('should uppercase the method', () => {
-  expect(oasToHar(oas, { path: '/', method: 'get' }).log.entries[0].request.method).toBe('GET');
+test('should uppercase the method', async () => {
+  const har = oasToHar(oas, { path: '/', method: 'get' });
+
+  expect(har.log.entries[0].request.method).toBe('GET');
+  await expect(har).toBeAValidHAR();
 });
 
 describe('url', () => {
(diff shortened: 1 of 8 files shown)


erunion:
What's the issue? I made some changes to HAR generation in the SDK work to pass Content-Type headers through so JSON objects are dumped in a snippets as objects, not stringified.

domharrington:
Sorry I should've been clearer. Form data values in the form are not coming through into the code sample!

image

erunion:
Tests will pass when https://github.com/readmeio/fetch-har/pull/67 is pulled into a new `fetch-har` release.

packages/oas-to-har/__tests__/index.test.js
```diff
@@ -1,4 +1,4 @@
-const querystring = require('querystring');
+const validate = require('har-validator');
```
erunion, line 1: FYI, most of the changes to this file were to rewrite some tests into using `it.each` to make them easier to maintain.

packages/oas-to-har/__tests__/index.test.js
```diff
@@ -7,16 +7,40 @@ const commonParameters = require('./__fixtures__/common-parameters');
 
 const oas = new Oas();
 
-test('should output a har format', () => {
-  expect(oasToHar(oas)).toStrictEqual({
+expect.extend({
+  toBeAValidHAR(har) {
```
domharrington, line 11: 🤯

packages/oas-to-har/__tests__/index.test.js
```diff
@@ -82,324 +110,200 @@ describe('parameters', () => {
...
-          },
-          {}
-        ).log.entries[0].request.url
-      ).toBe('https://example.com/param-path/id');
-    });
-
-    it('should use example if no value', () => {
-      expect(
-        oasToHar(oas, {
+    it.each([
```
domharrington, line 113: 👌

packages/oas-to-har/__tests__/index.test.js
```diff
@@ -1157,7 +1048,7 @@ describe('requestBody', () => {
             },
           },
         }).log.entries[0].request.postData.text
-      ).toBe(querystring.stringify({ a: 'value' }));
+      ).toBe(JSON.stringify({ a: 'value' }));
```
domharrington, line 1051: Is this correct? How come it's change from querystring to JSON?

erunion:
@domharrington Yeah it seems like it was a result of me adding setting `postData.mimeType` to `application/x-www-form-urlencoded` with the node-simple work in https://github.com/readmeio/api-explorer/pull/792/files#diff-528f99451eda35186e98d5f772a60fd1R182, and we didn't have sufficient tests for formData anywhere else.

packages/oas-to-har/__tests__/index.test.js
erunion: I'm tempted to rip this out into its own package.

Issue #89, opened 2018-01-03
Need to support cookie authentication?


domharrington:
https://github.com/readmeio/readme/pull/731

PR #817, opened 2020-07-01
feat: adding support for cookie auth

erunion:
## 🧰 What's being changed?

This adds support for being able to handle [cookie auth](https://swagger.io/docs/specification/authentication/cookie-authentication/) throughout our general Explorer authentication flows.

It functions identical to how `apiKey` works, just that if `in` is set to `cookie`, the payload will be sent as a cookie. 

Resolves https://github.com/readmeio/api-explorer/issues/89

## 🧪 Testing

I've added an addition to our `auth-types` example and support can be seen there:

Screen Shot 2020-07-01 at 1 09 05 AM

For comparison this is what it looks like against `master`:

Screen Shot 2020-07-01 at 1 11 35 AM


packages/oas-to-har/src/lib/configure-security.js
```diff
@@ -50,6 +50,11 @@ module.exports = function configureSecurity(oas, values, scheme) {
       }
 
       return harValue('headers', header);
+    } else if (security.in === 'cookie') {
+      return harValue('cookies', {
+        name: security.name,
+        value: values[scheme],
+      });
```
erunion, line 57: 🙆‍♂️

erunion:
`api` doesn't support this yet, but will with https://github.com/readmeio/api/issues/38.

Screen Shot 2020-07-01 at 1 23 43 AM

erunion:
Really not sure if I want to ever support reading from `document.cookie` when we have JWT readily available. But yeah Link stuff. 😬 

We should have a larger discussion about that some point.

domharrington:
I think if we had Link support we'd never need to read from document.cookie right?

erunion:
Kind of, but a (the?) problem with the Link spec right now is that it has no formal support for auth flows so there isn't quite a spec-friendly way to grab data from a response and re-use it in auth.

PR #820, opened 2020-07-01
enhancement(rdmd): more compat mode features!

### LINKED ISSUE #668: RDMD Compatibility Mode
Per the discussion [here](https://github.com/readmeio/api-explorer/pull/660#issuecomment-621908636), we'd like to add a mode to the RDMD engine that enables certain, otherwise non-compliant behaviors in the name of backwards compatibility! Here's a quick list of included features off the top of my head:

- [x] Block components in tables (#660)
- [x] Render magic headings as `<h1>`s (#820)
- [x] Script JS execution in Custom HTML blocks (#820) [](https://stackoverflow.com/a/35616264/1341949 "Executing script tags in dangerously set HTML")

rafegoldberg:
## 🧰 Changes

This PR rounds out the compatibility mode functionality and closes #668. Turning on compat mode is meant to mimic our old markdown processor's behavior, so it should:

- [x] Render magic heading blocks with `<h1>` tags.
- [x] Execute all scripts found in Custom HTML blocks.
- [x] Allow block-level markdown in tables (in #660.)

## 🧪 Testing

_ReadMe integration to come._


erunion:
Not psyched about `eval()` but since it's just in customer markdown that's self-XSS territory if it gets abused.

rafegoldberg:
@erunion strong agree, hence the decision to shove this in to compatibility mode, instead of running user scripts by default!

PR #828, opened 2020-07-12
ci: setting up semantic-release

erunion:
## 🧰 What's being changed?

Sets up a new [semantic-release](https://github.com/semantic-release/semantic-release) flow for automatically building releases from `next` into `master`.

**`next` is now the base branch of this repository.**


.github/workflows/release.yml
```diff
@@ -0,0 +1,70 @@
...
+      - name: Install packages
+        run: |
+          npm ci
+          npx lerna bootstrap --concurrency 2
+        env:
+          CI: true
+
+      # Version, build, tag, and publish the new release;
+      #
+      - name: Build new release
```
erunion, line 41: @rafegoldberg So the current flow for building a new Explorer release is to run:

```
npm run build
npm run deploy
npm run publish // this runs `npx lerna publish`
```

I guess for  the `build && deploy` stage that should be its own step before "Build new release", but any ideas on how or where `npx lerna publish` should be run to update the individual packages?

.github/workflows/release.yml
```diff
@@ -0,0 +1,70 @@
...
+        run: npm run release # see: .releaserc for config
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_ADMIN || secrets.GITHUB_TOKEN }} # auth to push release to remote repo
```
erunion, line 44: @rafegoldberg Where should I get these Github and NPM tokens from?

.github/workflows/release.yml
```diff
@@ -0,0 +1,70 @@
+name: Release
+
+on:
+  push:
+    branches:
+      - master
+      - next
```
erunion, line 7: @rafegoldberg I don't think we want to build releases against `next` here because of the amount of Dependabot updates that we get so it's probably safe to remove this and just build a new release whenever `next` is merged into `master`, right?

.github/workflows/release.yml
```diff
@@ -0,0 +1,70 @@
...
+          branch: next
+        continue-on-error: true
+
+      # Save the bundle as a GitHub workflow artifact.
+      #
+      - name: Save bundle
+        uses: actions/upload-artifact@v1
+        with:
+          name: Bundle
+          path: dist
```
erunion, line 69: This we have a bunch of different dist directries here, thoughts on which we should use for the artifact bundle? Should we just change this to `packages/`?

.releaserc
```diff
@@ -0,0 +1,146 @@
...
+      "@semantic-release/changelog",
+      {
+        "changelogTitle": "Changelog\n==="
+      }
+    ],
+    [
+      "@semantic-release/npm",
+      {
+        "npmPublish": true,
+        "tarballDir": "dist"
```
erunion, line 123: idk what to change this to since we're using Lerna.

.github/workflows/release.yml
rafegoldberg: Yeah, that's what we do on ReadMe as well.

.github/workflows/release.yml
rafegoldberg: Semantic release should automatically run the `npm publish` command at the end of the process!

.github/workflows/release.yml
rafegoldberg: You only really the need the `GITHUB_TOKEN` one. I used a personal access token, but in the long run we should probably generate dedicated tokens for this. (I think an org-level admin can generate access tokens tied to the organization, rather than an individual account.)

.github/workflows/release.yml
rafegoldberg: That's one option. We could also remove this step entirely, I think, since I believe GitHub artifacts the entire source on tag if no assets are specifically given.

.releaserc
```diff
@@ -0,0 +1,146 @@
+{
+  "branches": [
+    "master",
+    {
+      "name": "next",
+      "prerelease": true
+    }
+  ],
```
rafegoldberg, line 8: We should probably remove the prerelease channel, since we're only running the release workflow on master anyways...

```suggestion
  "branches": [
    "master"
  ],
```

.releaserc
```diff
@@ -0,0 +1,146 @@
...
+        "changelogTitle": "Changelog\n==="
+      }
+    ],
+    [
+      "@semantic-release/npm",
+      {
+        "npmPublish": true,
+        "tarballDir": "dist"
+      }
+    ],
```
rafegoldberg, line 125: So wrt/[this comment](https://github.com/readmeio/api-explorer/pull/828/files#r453374712), I think you might be able to get away with just removing the options and letting it run through the default lerna publishing flow?

```suggestion
    "@semantic-release/npm",
```

rafegoldberg: (I also spent a bunch of time reading through [this issue](https://github.com/semantic-release/semantic-release/issues/193) re: setting up semantic-release in a monorepo environment... I'm honestly still not quite clear on the problems this presents, having not spent much quality time with lerna. But it does seem like it'll may take a bit more thought to get this set up than I'd've expected. 😭)

.github/workflows/release.yml
rafegoldberg: We should also [update the release branches](https://github.com/readmeio/api-explorer/pull/828#discussion_r453998516).

.releaserc
rafegoldberg: Responded [here](https://github.com/readmeio/api-explorer/pull/828#discussion_r454000133).

PR #852, opened 2020-07-23
feat: upgrading @readme/react-jsonschema-form to v1.2.0

erunion:
## 🧰 What's being changed?

Updates our fork of RJSF to pull in the potential fix for circular refs in https://github.com/readmeio/react-jsonschema-form/pull/14.

## 🧪 Testing

We have a demo circular ref definition that's currently failing here: https://preview.readme.io/?selected=swagger-files%2Fcyclical-refs.json

Loading it up locally via http://localhost:9966/?selected=swagger-files%2Fcyclical-refs.json you'll now see:

Screen Shot 2020-07-23 at 3 10 17 PM

### DIFF
--- a/packages/api-explorer/package-lock.json
+++ b/packages/api-explorer/package-lock.json
@@ -415,9 +415,9 @@
       }
     },
     "@babel/runtime-corejs2": {
-      "version": "7.9.2",
-      "resolved": "https://registry.npmjs.org/@babel/runtime-corejs2/-/runtime-corejs2-7.9.2.tgz",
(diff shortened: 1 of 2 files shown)


erunion:
@dok This circular example has `FormDataBodyPart` as its request body, and that only gets circular because `MultiPart` references itself. We should probably still be able to handle rendering the rest of `FormDataBodyPart`, just not the `parent` property on it.

dok:
Thanks for baking this. Should we have some notation of when a schema form doesn't get rendered because of a circular ref?

erunion:
Yeah we should either show a message saying that the schema property is circular, or just not show "Body Params" at all if the form isn't going to end up rendering anything.

erunion:
We're good with this as it is, but should probably revisit at some point after the fix gets upstream to give users a slightly better experience.

PR #867, opened 2020-08-04
feat: adding support for file uploads

erunion:
This adds support for file uploads!

giphy

How though...

### ✂️ Code snippets

Within the Explorer, when we handle form puts within the `Params` component, we generate two versions of the HAR: one for the `CodeSample` component that renders code snippets to the user, and another for the "Try It" functionality that'll make a request within the browser.

With the `CodeSample` component, the code snippets we want for file uploads to be clean, which means that we'd like them to not contain the jumbled, incomprehensible, visually useless [data URL](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs) encode of the file and instead utilize the filename of the file being uploaded. Thankfully we have this data via the data URLs that RJSF generates for us so we do a couple things:

* Decode the data URL, when running within the `CodeSample` component and grab the file name.
* In `oas-to-har` we attach the file name, and if we have it a content type, into the HAR so it can be used in snippet genration.

This brings us to the first major problem, which is that [httpsnippet](https://www.npmjs.com/package/httpsnippet) didn't quite support `multipart/form-data` HARs. Well... it did, but only when running within a server environment. This is because `httpsnippet` utilizes a library called [form-data](https://www.npmjs.com/package/form-data) that aims to allow Node environments access to the browser's [FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData) object.

The problem with this though? The API for `form-data` doesn't 100% match `FormData`, and if `form-data` is loaded in the browser, like if it were compiled into an application via Webpack, it loads the native `FormData` API. Now this normally wouldn't be a problem, but `httpsnippet` was built against the Node version of the `form-data` API, and not the browser one. So in order to allow us to compile `multipart/form-data` snippets within the browser, we needed to update `httpsnippet` in order to detect where it's being run, and adopt its handling accordingly. You can see this code, and a heap of other file upload and consistency related fixes, here: https://github.com/readmeio/httpsnippet/pull/1

With this change, we can now generate `multipart/form-data` code snippets for the user that are clean and will work outside of the browser!

Screen Shot 2020-08-06 at 3 12 53 PM

Screen Shot 2020-08-06 at 3 12 57 PM

For comparison, this these snippets would look like if we weren't parsing out file names:

Screen Shot 2020-08-06 at 3 14 42 PM

That's a big yikes.

### 🏃 "Try It"

For the "Try It" functionality, since we don't have access to the original file through RJSF, just the encoded data URL, we're sending that along in our payloads.

Screen Shot 2020-08-06 at 3 17 31 PM

This should work just fine since data URLs contain the full contents of the file, and are properly encoded to tell the consuming server what it is.

### 🧪 Testing

I've added a couple new API definitions to the demo site to these these out with:

| Name | What it tests | Quirks |
| :--- | :--- | :--- |
| `file-upload-image` | This will test uploading an `image/png` file. | For describing payload bodies that don't contain parameters, the content needs to be stringified and used as `postData.text`. This unfortunately doesn't give us file metadata advantages as `postData.params` so we can't decode data URLs, obtain file names, and generate clean code snippets for these kinds of operations. |
| `file-upload-multipart` | This tests out a standard `multipart/form-data` operation that has an `orderId`, `userId`, and a `documentFile` in the payload. | None! 😅  |
| `file-upload-multiple` | `multipart/form-data` where files are uploaded as an array. | The HAR spec doesn't offer a clear path for supporting these kinds of operations, so instead we need to stringify the array of files and submit that as the data for the parameter. |

### 📓 Still to do
* [x] Fix handling of `multipart/form-data` where files are in an array. See the `file-upload-multiple` example.
* [ ] Make sure `fetch-har` is good with all these changes.
* [ ] Update API SDK code snippets to allow for cleaner file snippets. https://github.com/readmeio/api/issues/131
* [ ] Update API SDK code snippets to not dump in an unnecessary `multipart/form-data` and a boundary header in. https://github.com/readmeio/api/issues/130


erunion:
@domharrington @rafegoldberg The `file-upload-multiple` case doesn't work yet, but this is otherwise good for an early review.

packages/oas-to-har/src/index.js
```diff
@@ -187,66 +217,119 @@ module.exports = (
...
-        if (jsonTypes.length) {
-          // We have to clone the body object, otherwise the form
-          // will attempt to re-render with an object, which will
-          // cause it to error!
-          let cloned = removeUndefinedObjects(JSON.parse(JSON.stringify(formData.body)));
-          jsonTypes.forEach(prop => {
-            // Attempt to JSON parse each of the json properties
-            // if this errors, it'll just get caught and stringify it normally
-            cloned[prop] = JSON.parse(cloned[prop]);
+        const cleanBody = removeUndefinedObjects(JSON.parse(JSON.stringify(formData.body)));
```
dok, line 255: Since we're parsing/stringifying, should we wrap this in a try...catch block?

packages/oas-to-har/src/index.js
```diff
@@ -187,66 +217,119 @@ module.exports = (
...
+          // them before submit, otherwise they'll be escaped instead of actual objects.
+          const jsonTypes = Object.keys(schema.schema.properties).filter(
+            key => schema.schema.properties[key].format === 'json'
+          );
+
+          if (jsonTypes.length) {
+            // We have to clone the body object, otherwise the form will attempt to re-render with an object, which will
+            // cause it to error!
+            try {
+              let cloned = removeUndefinedObjects(JSON.parse(JSON.stringify(formData.body)));
```
dok, line 303: Can we re-use `cleanBody` from above?

packages/oas-to-har/src/index.js
```diff
@@ -40,13 +41,42 @@ function isPrimitive(val) {
...
+function stringify(json) {
+  return JSON.stringify(removeUndefinedObjects(typeof json.RAW_BODY !== 'undefined' ? json.RAW_BODY : json));
+}
+
+function getNameFromDataUrlType(type) {
+  const properties = type.split(';').filter(param => {
+    return param.split('=')[0] === 'name';
+  });
+
+  return properties.length !== 1 ? 'unknown' : properties[0].split('=')[1];
```
dok, line 53: How does this line work? Wouldn't the ternary statement always evaluate to a Boolean value?

packages/oas-to-har/src/index.js
erunion: Not easily. I'd need to wrap 100+ lines or so within a try-catch and this code is a lot easier to parse without that.

erunion: Eh, maybe actually it'll look okay.

gkoberger:
image

packages/oas-to-har/src/index.js
```diff
@@ -14,11 +15,17 @@ function formatter(values, param, type, onlyIfExists) {
     return undefined;
   }
 
-  if (param.required && param.example) {
-    return param.example;
+  if (param.required && param.schema && param.schema.default) {
```
erunion, line 18: Making this change because we shouldn't be using examples in compiled HAR files we make requests with. If there's a value set in the `default` we should use that instead. Examples should only be for reference, not actual use.

packages/oas-to-har/src/index.js
```diff
@@ -14,11 +15,17 @@ function formatter(values, param, type, onlyIfExists) {
...
   }
 
-  return param.name;
+  // If we don't have any values for the path parameter, just use the name of the parameter as the value so we don't
+  // try try to build a URL to something like `https://example.com/undefined`.
+  if (type === 'path') {
+    return param.name;
+  }
+
+  return undefined;
```
erunion, line 28: Made this change because for our own API definition, if a user didn't specify an `x-readme-version` header, we'd send a header that evaluated to `x-readme-version: 'x-readme-version'` which isn't right at all.

PR #895, opened 2020-08-24
fix: creating a custom DateTimeWidget for datetime-local inputs

erunion:
## 🧰 What's being changed?

`<input type="datetime-local">` is not fully supported by all browsers, and when we pull the stock `DateTimeWidget` from RJSF in, browsers that don't support the native input run into a weird issue where the input acts as if it's disabled.

The reason it's happening is because RJSF does some unnecessary UTC conversion work this input (as far as I can tell at least from reading the MDN docs for browsers that natively support it), and when that work happens on browsers that can't handle it like Firefox they don't surface a TZ date `onChange`. When that happens, RJSF just returns an empty string, which in turn undoes any input that the user entered rendering the input effectively disabled.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local

## 🧪 Testing

Open up http://localhost:9966/?selected=swagger-files%2Ftypes.json in Chrome and Firefox and scroll till you see the following inputs:

Screen Shot 2020-08-24 at 4 34 24 PM

In Chrome they should look and act like the above, whereas in Firefox the `date-time` input should downgrade to a stock text input.

Screen Shot 2020-08-24 at 4 34 24 PM

### DIFF
--- a/packages/api-explorer/src/Params.jsx
+++ b/packages/api-explorer/src/Params.jsx
@@ -4,7 +4,7 @@ const Form = require('@readme/react-jsonschema-form').default;
 const slug = require('lodash.kebabcase');
 
 const {
-  // DateTimeWidget,
+  DateWidget,
   PasswordWidget,
   TextWidget,
   UpDownWidget,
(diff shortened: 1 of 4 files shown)


packages/api-explorer/src/Params.jsx
```diff
@@ -133,6 +131,7 @@ class Params extends React.Component {
 Params.propTypes = {
   ArrayField: PropTypes.func.isRequired,
   BaseInput: PropTypes.func.isRequired,
+  DateTimeWidget: PropTypes.func.isRequired,
```
darrenyong, line 134: what does [prop-types](https://www.npmjs.com/package/prop-types) do? it says it checks for React props, but the [usage section](https://www.npmjs.com/package/prop-types#usage) doesn’t seem too clear on what it’s actually doing. in the [React Docs](https://reactjs.org/docs/typechecking-with-proptypes.html), it looks like it’s a kind of validator that makes sure the props that are being passed are of correct type?

erunion: Yep that's what it does. Ensures that when you pass one of these properties to a React component that it's of the type specified here. If it isn't, React will throw a soft error in the console.

darrenyong: oops. accidentally messed up the review requesters 😓

PR #930, opened 2020-09-01
chore: adding jsx transpilation to @readme/variable

gratcliff:
## 🧰 What's being changed?
Adding a webpack source to transpile our jsx into universal consumable output.

Screen Shot 2020-09-01 at 3 29 47 PM


packages/variable/webpack.config.js
```diff
@@ -0,0 +1,31 @@
...
+module.exports = {
+  entry: ['./index.jsx'],
+  module: {
+    rules: [
+      {
+        test: /\.js(x?)$/,
+        use: {
+          loader: 'babel-loader',
+          options: {
+            extends: './.babelrc',
```
erunion, line 13: You should be able to use the root Babel config instead

```suggestion
            extends: '../../.babelrc',
```

Issue #911, opened 2020-08-31
Allow custom extensions to be used on an operation basis


erunion:
Right now if a customer wants to disable the "Try It" functionality on a single operation they need to pull that operation into its own API Definition and upload that. This is just silly. We should allow customers to use our custom OAS extensions on an operation-by-operation basis, allowing them to override settings that are set at the global level.

For example, if `"x-explorer-enabled": true,` is set at the root level of the API definition, the API Explorer will be enabled for all operations in the API, but if `"x-explorer-enabled": true,` is set inside `paths./pet.post` to `"x-explorer-enabled": false,`, it'll be disabled for just that operation.

PR #940, opened 2020-09-08
feat: operation level extensions

darrenyong:
## 🧰 What's being changed?

Currently, our [Swagger extensions](https://docs.readme.com/guides/docs/openapi-extensions) are only allowed on a file basis. I've added checks to allow customers to use our Swagger extensions on an operation basis. Rather than using the extensions off the oas file, we'll first check to see if it exists on the operation level. If not, we'll just use the defaults.

Resolves https://github.com/readmeio/api-explorer/issues/911

## 🧪 Testing
Add any extension on an operation level (same level as the `summary`, `responses`, and `requestBody`). There should be a difference between the different endpoints now!

In this image, the `/GET` doesn't have the explorer enabled, but the `/POST` does. It also only has the `cURL` language sample enabled while the `/POST` has them all!
image


darrenyong:
Still have some to-do's:
- [ ] `x-proxy-enabled` extension
- [ ] `x-proxy-enabled` extension
- [ ] `x-proxy-enabled` extension
- [ ] write tests to ensure we always check operation level

packages/api-explorer/src/Doc.jsx
```diff
@@ -43,7 +43,14 @@ class Doc extends React.Component {
...
     this.waypointEntered = this.waypointEntered.bind(this);
-    this.Params = createParams(this.oas);
+
+    const operation = this.getOperation()
+    const operationExtension =
+      operation[extensions.EXPLORER_ENABLED] !== undefined
+        ? operation
+        : this.oas
+
+    this.Params = createParams(operationExtension);
```
erunion, line 53: Since we're supplying both `this.oas` and `this.getOperation()` into `this.Params`, this call to `createParams` can remain at `this.oas`.

And to help clean up all the different parameter widgets, what you can probably do instead is create a new prop on the `Params` component called `explorerEnabled`. Default it to `true`, and then here in `Doc` when we do `<this.Params ... />`, supply that prop with your logic here. Then in the different widget components, you can clean it up to just look at `explorerEnabled`.

Doing this should allow you to clean up a lot all of the repeated `const explorerEnabled = oas[extensions.EXPLORER_ENABLED];` code that we've got there.

darrenyong:
> Still have some to-do's:
> 
> * [ ]  `x-proxy-enabled` extension
> * [ ]  `x-headers` extension
> * [ ]  `x-send-defaults` extension
> * [ ]  write tests to ensure we always check operation level

@erunion, we won't need to do this for `x-send-defaults` right? it looks like it's been deprecated now.

erunion:
Yeah don't worry about `x-send-defaults`.

packages/api-explorer/src/Params.jsx
```diff
@@ -150,17 +151,17 @@ Params.defaultProps = {
...
   // eslint-disable-next-line react/display-name
   return props => {
+    const ArrayField = createArrayField(oas, props.explorerEnabled);
+    const BaseInput = createBaseInput(oas, props.explorerEnabled);
+    const DateTimeWidget = createDateTimeWidget(oas, props.explorerEnabled);
+    const FileWidget = createFileWidget(oas, props.explorerEnabled);
+    const SchemaField = createSchemaField();
+    const SelectWidget = createSelectWidget(oas, props.explorerEnabled);
+    const TextareaWidget = createTextareaWidget(oas, props.explorerEnabled);
+    const URLWidget = createURLWidget(oas, props.explorerEnabled);
```
darrenyong, line 163: @erunion not sure why it's saying I'm missing validation for these props :(

image

but I have them on the [Params Component](https://github.com/readmeio/api-explorer/pull/940/files#diff-3c697e1c364b5422881949b233f9aa3cR135) and on each of the different widget components:
* [Response](https://github.com/readmeio/api-explorer/pull/940/files#diff-3603347a29fff625128f5ffc722c3759R75)
* [Response Example](https://github.com/readmeio/api-explorer/pull/940/files#diff-99031c112c99dd4e1b4a1bda546ffc7bR275)
* [Array Field](https://github.com/readmeio/api-explorer/pull/940/files#diff-a96510858c4571c0e832e75c83da461bR49)
and so on

erunion: It might be because it's coming through as just an object on line 155 and ESLint doesn't see it as being part of the `Params` component props. To fix it you can probably save `props.explorerEnabled` to a local `explorerEnabled` constant in `createParams`, silence the error with `// eslint-disable-next-line react/prop-types` and then pass `explorerEnabled` into these creation calls.

darrenyong: did some refactoring. so close to passing tests 😞

darrenyong:
I HAVE CONQUERED THE TESTS!
IMG_0177

Time to write my own 😐

darrenyong:
ready for review 😄

darrenyong:
so Greg found an issue where selecting a code language that didn't exist on an endpoint due to the `x-samples-languages` extension would cause that endpoint to show the code sample of the selected language. hopefully this picture is more clear:
image
basically, I selected JavaScript as the language, but my `GET` endpoint is displaying the JavaScript language even though it's not "allowed" in `x-samples-languages`.

have a fix in place, but will probably need some help on CSS 😂 can chat when you're back!

gkoberger:
😍

erunion:
Spent a bit of time looking into fixing the bug where if you have different languages between operations the rendering will be weird and it's not going to be an easy fix because the `setLanguage` callback sets `is-lang-{currentLanguage}` on the global body of the Explorer. Reworking that to be independent to each operation is going to be a problem because we'll still need to account for the case when you toggle between a code sample in one operation, if that language is present in others it should show it there too.

So yeah, I'm punting that bug for now.

erunion:
I'm going to revert the operation-specific support for `x-samples-languages` in this PR until https://github.com/readmeio/api-explorer/issues/965 is resolved because I don't want us to ship that issue out.

PR #941, opened 2020-09-10
docs: update installation instructions

kanadgupta:
## 🧰 What's being changed?

Removed a now unnecessary line since we do this on the postinstall now, right? https://github.com/readmeio/api-explorer/blob/dacbe08d5c3c4cb0aa9a6f19b2af95a9ba20380c/package.json#L13

(also we should update the husky rules for this repo so I can add permalinks in commit messages, since those are typically >100 characters)

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,6 @@ This repo consists of the following npm modules:
 git clone git@github.com:readmeio/api-explorer.git
 cd api-explorer
 npm install
-npx lerna bootstrap
 ```
 
 ## Running the tests


erunion (approved):

erunion:
You should be able to add links in commit messages, just in the commit description/body.

kanadgupta:
> You should be able to add links in commit messages, just in the commit description/body.

It wasn't working in the body either 😞

erunion:
🤔 will take a look

erunion:
fixed in https://github.com/readmeio/api-explorer/commit/94f47988a4b18c6a16b1ce06168bd1cdf70e9bd4

PR #949, opened 2020-09-15
chore: remove syntax-highlighter from api-explorer

julshotal:
## 🧰 What's being changed?

- removed `syntax-highlighter` from the `api-explorer`
- changed `uppercase` require to use object deconstruction & updated file path
- added `.default` where `syntax-highlighter` was brought in (needed otherwise module imports incorrectly)
- updated packages to the 9.0.0 release of `syntax-highlighter`

## 🧪 Testing

run locally using `npm start`


erunion (approved):

erunion:
@gratcliff This looks good to me, but if it's ok with you I'm going to create a release just for it after https://github.com/readmeio/api-explorer/pull/950 is merged in so if there's any problem DOM changes with this, yall can safely rollback next week without undoing the cyclical fix in https://github.com/readmeio/api-explorer/pull/951.

gratcliff:
> @gratcliff This looks good to me, but if it's ok with you I'm going to create a release just for it after #950 is merged in so if there's any problem DOM changes with this, yall can safely rollback next week without undoing the cyclical fix in #951.

@erunion I think that makes complete sense! Thanks for giving this a look over, and in your help in diagnosing some of those dependency issues!

PR #950, opened 2020-09-17
fix: reverting `date` and `date-time` changes in #895

erunion:
## 🧰 What's being changed?

This reverts the changes to `date` and `date-time` that were introduced in https://github.com/readmeio/api-explorer/pull/895.

The `datetime-local` input isn't widely supported, and doesn't properly support datetimes so we're reverting and treating them, and `date` inputs, as simple strings for the indefinite future.

### DIFF
--- a/example/swagger-files/types.json
+++ b/example/swagger-files/types.json
@@ -97,6 +97,7 @@
                     "format": "date"
                   },
                   "string (format: date-time)": {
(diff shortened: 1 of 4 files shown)


darrenyong (approved):
Hey I actually understand what's happening 😂 
Tested it in the API Explorer to make sure it's string type now!

Looks good to me!

PR #951, opened 2020-09-18
chore(deps): upgrading @readme/react-jsonschema-form to 1.2.1

erunion:
## 🧰 What's being changed?

This upgrades `@readme/react-jsonschema-form` to v1.2.1 to resolve an issue introduced in v1.2.0 where we stopped processing schemas that were cyclical, but ones that were so in a safe way that wouldn't cause infinite recursion.

Check https://github.com/readmeio/react-jsonschema-form/pull/15 for more details on the problem and the actual fix.

## 🧪 Testing

You can see an example of this in the `/not-quite-cyclical` example I added into the `cyclical-refs` example definition, which you can also see here: http://bin.readme.com/s/5f63f9352fd27d002432212a

Before our fix, this operation wouldn't have any body parameters rendered:

Screen Shot 2020-09-17 at 5 03 20 PM

But now:

Screen Shot 2020-09-17 at 5 03 44 PM

### DIFF
--- a/example/swagger-files/cyclical-refs.json
+++ b/example/swagger-files/cyclical-refs.json
@@ -1,15 +1,17 @@
 {
   "openapi": "3.0.1",
   "info": {
-    "title": "Document Resource",
-    "description": "Use this for adding, updating or removing documents.",
+    "title": "Cyclical examples",
     "version": "1.0.0"
   },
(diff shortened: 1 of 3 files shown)


dok (approved):
LGTM if the build passes!

PR #961, opened 2020-09-28
fix: parse example.response.value.$ref rather than render as plain text

darrenyong:
## 🧰 What's being changed?
 | 🎫 [Ticket](https://app.asana.com/0/1178132843162889/1148431658605906)|
:---:

If a customer has a `$ref` in their example responses, we currently render these as plain text. This change would find the schema definition for the `$ref` and show the schema instead! 

We're currently only checking if `example.response` exists — if it does, we're returning the `value`. Since the `$ref` can be contained in the `value` field, I've added logic to check for the existence of `$ref` first. If it exists, we're going to call `findSchemaDefinition` on the `$ref` and the oas file.

## 🧪 Testing
1. Find a file that's using a `$ref` in their response examples (or use this one [example with ref.json.zip](https://github.com/readmeio/api-explorer/files/5290347/example.with.ref.json.zip))
2. Pull it into preview.readme.io or bin.readme.com
3. Notice that the endpoint will display the `$ref` as plain text
4. Check on this branch with the same file.
5. It should find the `$ref` and render it correctly.


darrenyong:
This currently will only work if the structure is `examples.response.value`. It won't work if the response is named something custom to be `examples.{custom-name}.value` as we're only checking for `response`.

erunion:
@darrenyong Can you add the example w/ ref spec into `example/swagger-files` somewhere?

erunion (approved):
Not going to worry about tests right now because all this response example work is going to get rewritten soon anyways when we tackle our `example` vs `examples` issue.

PR #962, opened 2020-10-01
fix: render 401 JSON resposes if the request was not initiated via OAuth

darrenyong:
## 🧰 What's being changed?
When we get a 401 response, if the response isn't initiated via OAuth, we just show:
> You couldn't be authenticated

Now, I added some logic to render the responseBody if there isn't OAuth. The logic I have feels a little bad, though. Definitely want to chat about it when you get back!

## 🧪 Testing

Provide as much information as you can on how to test what you've done.

### DIFF
--- a/packages/api-explorer/src/ResponseBody.jsx
+++ b/packages/api-explorer/src/ResponseBody.jsx
@@ -59,11 +59,31 @@ function hasOauth(oauth) {
   );
 }
 
-function Unauthorized({ isOauth, oauth }) {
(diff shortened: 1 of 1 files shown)


darrenyong:
after doing some thinking, I was wondering if it's better to add the 401 response code/OAuth checking logic in the `ResponseBody` functional component. that way, we can decide whether to use the `ReactJson` component or the Unauthorized component.

erunion:
We're handling this in the redesign so I'm going to close this out

PR #969, opened 2020-10-09
feat: adding support for the `example` property on responses

### LINKED ISSUE #922: Response examples requires a plurality that differs from the spec
The OpenAPI Spec defines singular `example` field for 1 example response, and plural `examples` field for multiple response examples: https://swagger.io/docs/specification/adding-examples/.

Currently, we appear to only process the `examples` plural, regardless of how many examples are provided. A singular example is rendered when under "examples" field. 

However, we don't render anything when given correct singular `example`, for instance see first endpoint: http://bin.readme.com/s/5ec5406bb2c7130024f5689f. This example is rendered https://editor.swagger.io/ but not in the explorer.

erunion:
## 🧰 What's being changed?

This adds long requested support for the `example` property on responses. 

There's some documentation on the property on https://swagger.io/docs/specification/adding-examples/#request-response-bodies, but it works the same as our existing `examples` support, just that all data attached to the `example` is treated as an example for that response.

I should note that the OAS spec, despite having an [`examples` component](https://swagger.io/docs/specification/components/) does not support attaching `$ref` to `example` properties. We currently follow the spec in this regard so if a `$ref` is in an example like in the following, we do **not** resovle it:

```yaml
paths:
  /example:
    get:
      responses:
        200:
          description: OK
          content:
            application/json:
              example:
                $ref: "#/components/examples/user"
```

Check out the [Media Type Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.1.0.md#mediaTypeObject) part of the spec for some more details on this.

Resolves #922

## 🧪 Testing

You can see this feature deficiency here: http://bin.readme.com/s/5f80b9a6fcbbd70024cfc4ea

Despite having an `example` object on these two responses, nothing is rendered in the examples box:

Screen Shot 2020-10-09 at 12 31 59 PM

With this change:

Screen Shot 2020-10-09 at 12 30 20 PM


darrenyong:
Looks good! To my understanding, adding a `$ref` under `example` will still render it as plain-text, correct? The person will need to add it under `examples.response.value` for it to work?

erunion:
@darrenyong Putting a `$ref` in `example` will render the example out as the raw ref instead of resolving it, yeah:

```json
{
  "$ref": "#/components/..."
}
```

`$ref` in an example is only supported in `examples.*.value`, yeah.

darrenyong (approved):

PR #982, opened 2020-10-13
style: update the oas form package to abide by our code standards

erunion:


erunion:
Tests pass locally but not on CI because the package isn't properly set up yet. Merging anyway.

PR #987, opened 2020-10-15
fix: bugs with processing examples

erunion:
## 🧰 What's being changed?

* [x] Fixes a bug where if an example value in the `examples` property on a response schema was a string we'd error out because we were attempting to look for a `$ref` in it.
* [x] Fixes a bug where non-JSON content types would be initially attempted to be fed into `react-json-view` before defaulting to plain text. This would break on JSON-encoded strings like `'"app1,app2"'` that `JSON.parse()` without errors, but since it's still not actually JSON `react-json-view` would show an error in its place.

Screen Shot 2020-10-15 at 10 57 11 AM


darrenyong:
Looks good! Checked out the OAS files that were affected and they seem to be rendering correctly now!

darrenyong (approved):

PR #999, opened 2020-10-20
fix: updating the ResponseSchemaBody for the async flattenSchema changes

erunion:
## 🧰 What's being changed?

With https://github.com/readmeio/oas/pull/292, `oas/tooling/lib/flattenSchema` is now an async library to enable us to properly handle circular `$ref` pointers when flattening a response in the `ResponseSchemaBody` component.

This change updates the `ResponseSchemaBody` component to be compatible with the changes incoming in https://github.com/readmeio/oas/pull/292.

## 🧪 Testing

Here's an example of a circular ref'd schema that's currently failing when we're flattening the response schema: http://bin.readme.com/s/5f8f6eec31f4240024e73488

### DIFF
--- a/packages/api-explorer/package.json
+++ b/packages/api-explorer/package.json
@@ -17,7 +17,7 @@
     "fetch-har": "^4.0.2",
     "js-cookie": "^2.1.4",
     "lodash.kebabcase": "^4.1.1",
-    "oas": "^4.0.0",
+    "oas": "^4.1.0",
     "prop-types": "^15.7.2",
     "react": "^16.4.2",
     "react-copy-to-clipboard": "^5.0.1",
--- a/packages/api-explorer/package.json
+++ b/packages/api-explorer/package.json
@@ -17,7 +17,7 @@
     "fetch-har": "^4.0.2",
     "js-cookie": "^2.1.4",
     "lodash.kebabcase": "^4.1.1",
-    "oas": "^4.1.0",
+    "oas": "^5.0.0",
     "prop-types": "^15.7.2",
     "react": "^16.4.2",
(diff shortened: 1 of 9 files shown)


packages/api-explorer/package.json
```diff
@@ -17,7 +17,7 @@
     "fetch-har": "^4.0.2",
     "js-cookie": "^2.1.4",
     "lodash.kebabcase": "^4.1.1",
-    "oas": "^4.0.0",
+    "oas": "^4.1.0",
```
erunion, line 20: I messed up and tagged this breaking change within `oas` to 4.1.0 instead of 5.0.0 so this PR might not work locally until I update this.

packages/api-explorer/src/ResponseSchemaBody.jsx
```diff
@@ -22,47 +22,73 @@ function getDescriptionMarkdown(useNewMarkdownEngine, description) {
...
+      });
+    });
+  }
+
+  render() {
+    const { schema, useNewMarkdownEngine } = this.props;
+    const { flattenedSchema } = this.state;
+
+    let rows = [];
+    if (flattenedSchema) {
```
gratcliff, line 49: Are we sure this'll always be iterable?

gratcliff (commented):
Looks good overall! Just one question below:

packages/api-explorer/src/ResponseSchemaBody.jsx
erunion: Yeah `flattenedArray` on line 39 where this state prop is set will always return an array.

darrenyong (approved):
Tested with the spec and renders without an error!

PR #1005, opened 2020-10-22
chore: pinning react and react-dom across every dependency

erunion:
## 🧰 What's being changed?

We're running into some React hook hell when running the latest version of the explorer with tutorials integration in ReadMe with tutorials present on a project. We *think* it's because the versions of React that the explorer, ReadMe, the UI library, and `@readme/syntax-highlighter` aren't all the same so what I'm doing here is:

* [x] Moving `react` and `react-dom` over to being dev dependencies, pinning them at the latest of the v16 series.
* [x] Adding the same dependencies, at `16.x` into being `peerDependencies`.
  * https://stackoverflow.com/questions/30451556/what-is-the-correct-way-of-adding-a-dependency-to-react-in-your-package-json-for
* [x] Upgrading `@readme/syntax-highlighter` across the explorer and `oas-to-snippet` to 10.1.0 which does the same peerdep work https://github.com/readmeio/syntax-highlighter/pull/24
  * The legacy Markdown engine `markdown-magic` is being excluded from this because upgrading is changing `span` tags to `div` and inserting empty `<code></code>` elements where there weren't before. Since its the legacy engine, it should be fine? Hopefully that not being the same as everything else isn't the linchpin in all of this.
* [x] Yakshaving every package file up a bit so they are all mostly identical.

I'm also extracting `react` and `react-dom` from being compiled into the explorer dist. Since the explorer is a reusable React component, it's always going to be used in an environment where React is present so we don't need to bloat up its bundle with another copy of React. The result is some **wild** drops in dist sizes:

* `@readme/api-explorer`: 5,199,205K to 4,644,622K. **11% reduction** 👀 
* `@readme/api-logs`: 33,273K to 25,863K
* `@readme/variable`: 15,266K to 7,919K


packages/api-explorer/package.json
```diff
@@ -11,49 +27,38 @@
...
+    "@readme/ui": "^1.10.2",
     "@testing-library/dom": "^7.26.3",
     "css-loader": "^4.3.0",
     "eslint": "^7.0.0",
     "jest": "^26.0.1",
     "jsinspect": "^0.12.6",
     "node-fetch": "^2.0.0-alpha.9",
     "prettier": "^2.0.1",
+    "react": "^16.14.0",
+    "react-dom": "^16.14.0",
```
rafegoldberg, line 61: Do we need these in dev deps just so they get installed when we `npm i`? If so, soon this won't be necessary (once we manage to bump up to [NPM 7](https://github.blog/2020-10-13-presenting-v7-0-0-of-the-npm-cli/)!)

rafegoldberg (approved):
This is great. Changes look okay to me. Have you tried running an integration test against ReadMe proper?

gratcliff (approved):

packages/api-explorer/package.json
erunion: @rafegoldberg yeah. once we're on npm 7 and they're defined as peerdeps we can remove them as being devdeps

PR #1017, opened 2020-10-29
feat: hide "suggest edits" button if there's no body

domharrington:
## 🧰 What's being changed?

Hide "suggest edits" button if there's no body

## 🧪 Testing

Check http://localhost:9966/


erunion (approved):

domharrington:
Dont think we wanna do this.

PR #1039, opened 2020-11-09
style: making the response example section scrollable

erunion:
## 🧰 What's being changed?

The new response example work in https://github.com/readmeio/api-explorer/pull/1027 introduced a new problem of large response schemas potentially blowing up the length of the page.

## 🧪 Testing

ezgif com-gif-maker


rafegoldberg (approved):
LGTM

PR #1050, opened 2020-11-17
feat: request body raw/json mode editor

erunion:
## 🧰 What's being changed?

This adds a new JSON tab on `application/json`-compatible request bodies to allow users to edit the raw JSON contents of their request.

Screen Shot 2020-11-18 at 5 42 22 PM

The initial contents of this editor is populated from the first example we can grab off the request. If we can't grab an example, we'll default to an empty JSON object.

**Syncing this data between the form inputs is not currently supported.**

* [x] JSON editor is enabled via a `enableRequestBodyJsonEditor` prop on the explorer.
* [x] Invalid JSON in the editor should throw an error message and not update the code samples.
* [x] Buttons available beneath the editor and in a new modal next to "Try It" to reset the JSON editor back to the original state.
* [x] Styles
  * [x] Changes do not conflict with the explorer when this editor is not available.
* [x] Tests

## 🧪 Testing

Load up the test server and check the petstore example. You should be able to:

* Tab to the JSON editor.
* Have the JSON editor prefilled with an example request.
* Edit the contents of the JSON editor with different data.
* Input invalid JSON and receive an error message at the bottom of the editor and up top at the submit button.
* Be able to reset the JSON editor at any time to the original state it was in (with the example request).
* Tab back to the Form side and have its original state intact.


packages/api-explorer/src/CodeSample.jsx
```diff
@@ -68,7 +68,7 @@ class CodeSample extends React.Component {
               <div key={key} style={{ display: selected ? 'block' : 'none' }}>
                 <CopyCode key={`copy-${key}`} code={example.code} />
                 <pre className="tomorrow-night tabber-body" style={{ display: selected ? 'block' : '' }}>
-                  {syntaxHighlighter(example.code, example.language, { dark: true })}
+                  {syntaxHighlighter(example.code, example.language, { customTheme: 'tomorrow-night' })}
```
erunion, line 71: `@readme/syntax-highlighter` no longer natively supports the `tomorrow-night` theme that we use for code samples so we're passing this in. CSS for the theme is already being loaded within the explorer and our Hub2 bundle in prod so we don't need to worry about that.

packages/api-explorer/src/Doc.jsx
```diff
@@ -25,44 +26,161 @@ const { Operation } = Oas;
 const parseResponse = require('./lib/parse-response');
 const Content = require('./block-types/Content');
 
-const pkg = require('../package.json');
+const { version: packageVersion } = require('../package.json');
```
erunion, line 29: Hopefully with this Webpack will stop adding our entire `package.json` into the dist. 😬

packages/api-explorer/src/Doc.jsx
```diff
@@ -352,7 +528,7 @@ class Doc extends React.Component {
       if (lazy) {
         return (
           <Waypoint bottomOffset="-1%" fireOnRapidScroll={false} onEnter={this.waypointEntered}>
-            {this.state.showEndpoint && this.renderEndpoint()}
+            {this.state.showEndpoint ? this.renderEndpoint() : <div />}
```
erunion, line 531: In test environments `react-waypoint` really doesn't like when we pass a bool as a child.

Dashron (approved):
Code looks good, had a question though about the feature. 

Should the form data sync with the JSON? 

e.g. if I put 12345 in the form's ID field, should the JSON initialize with the ID? Should it work the other direction too?

Not sure if this is an easy fix or not, but it would be pretty pleasant UX.

erunion:
@Dashron We decided for the MVP that data wouldn't sync between the form and json tabs because we'll need to introduce JSON schema validation to prevent people from turning object requests into arrays and breaking the form (which is very brittle).

packages/api-explorer/src/Doc.jsx
```diff
@@ -25,44 +26,178 @@ const { Operation } = Oas;
...
   }
 
+  onJsonChange(rawData) {
+    this.setState(previousState => {
+      let data;
+      try {
+        data = JSON.parse(rawData);
+
+        return {
+          dirty: true,
```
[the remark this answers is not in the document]
erunion: actually we might need to still split up the dirty flag and let the form and json editor maintain their own dirty states because right now if you edit the json tab and switch over to the form it'll be dirty despite nothing being changed in it.

Dashron:
Def makes sense. 👍

runnabro:
The button is still in the invalid state if the form request is good when:
1. Go to the json tab
2. Make it invalid
3. Go back to the form tab

erunion:
@runnabro @Dashron I just pushed up some changes to keep an independent dirty state for the form and JSON tabs to fix the issue of a dirty state being carried across.

Dashron (approved):

PR #1053, opened 2020-11-18
feat(apiexplorer): support passing function to handle group change 

dpim:
## 🧰 What's being changed?
- Add support for passing an optional function to handle group change via props (which is called in addition to existing, built-in behavior). 

This function would receive the new group member id whenever onAuthGroupChange() is invoked.

## 🧪 Testing
- Manually verified

### DIFF
--- a/packages/api-explorer/src/index.jsx
+++ b/packages/api-explorer/src/index.jsx
@@ -209,6 +209,11 @@ class ApiExplorer extends React.Component {
     const NewGlossaryTermsContext = utils.GlossaryContext;
     /* eslint-enable global-require */
 
+    const onAuthGroupChange = auth => {
+      if (this.props && this.props.onAuthGroupChange) this.props.onAuthGroupChange(auth);
+      return this.onAuthGroupChange(auth);
+    };
+
     return (
       <div className={`is-lang-${this.state.language}`}>
         <div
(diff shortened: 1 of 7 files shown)


packages/api-explorer/src/index.jsx
```diff
@@ -209,6 +209,11 @@ class ApiExplorer extends React.Component {
     const NewGlossaryTermsContext = utils.GlossaryContext;
     /* eslint-enable global-require */
 
+    const onAuthGroupChange = auth => {
+      if (this.props && this.props.onAuthGroupChange) this.props.onAuthGroupChange(auth);
```
gratcliff, line 213: Might want to check if this is `if (this.props && typeof this.props.onAuthGroupChange === 'function')` to continue `proptype` restrictions and ensure it's callable.

packages/api-explorer/src/index.jsx
```diff
@@ -209,6 +209,11 @@ class ApiExplorer extends React.Component {
     const NewGlossaryTermsContext = utils.GlossaryContext;
     /* eslint-enable global-require */
 
+    const onAuthGroupChange = auth => {
```
erunion, line 212: Instead of doing this here, can you move the following line into the top of the `onAuthGroupChange` function so this supplementary behavior is contained in there? And then in the `onAuthGroupChange` call on line 251 that can remain as `this.onAuthGroupChange`.

gratcliff, line 212: Since you're calling a class method `onAuthGroupChange` of the same name, for consistency and maintainability it might make sense to consolidate prop call inside `this.onAuthGroupChange()`. 

On 117:
```
 if (this.props && this.props.onAuthGroupChange) return this.props.onAuthGroupChange(group);
```

gratcliff (commented):
Looks fine to me! Couple nitpick comments below. Happy to discuss if you'd like.

erunion (changes requested):

packages/api-explorer/src/index.jsx
erunion: Also shouldn't need to verify that `this.props` exists since that's always present.

PR #1056, opened 2020-11-19
chore(oas-to-snippet): replace stringified request body with JSON.stringify and object

Dashron:
replace stringified request body with JSON.stringify and object

## 🧰 What's being changed?

The request body in the node example must be a string. Originally we were translating the JSON object to a string and including that. A JSON.stringified object is difficult to read, so this PR keeps the object literal and wraps it in a JSON.stringify call in the example.

See https://app.asana.com/0/1199087650851461/1198986894847661 for more discussion and example pictures.

## 🧪 Testing

Go to the petstore example, and the "add a new pet"  API call. Start typing in the "category" request body form below the example, and you should see the example automatically update with a request body that is an object literal wrapped in JSON.stringify.

### DIFF
--- a/packages/oas-to-snippet/src/index.js
+++ b/packages/oas-to-snippet/src/index.js
@@ -44,12 +44,6 @@ module.exports = (oas, operation, values, auth, lang, oasUrl) => {
     };
   }
 
-  if (lang === 'node') {
-    language.httpsnippet[2] = {
-      useObjectBody: true,
-    };
-  }
-
   return {
     code: snippet.convert(...language.httpsnippet),
     highlightMode: language.highlight,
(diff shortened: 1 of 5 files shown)


packages/oas-to-snippet/src/index.js
```diff
@@ -44,6 +44,12 @@ module.exports = (oas, operation, values, auth, lang, oasUrl) => {
     };
   }
 
+  if (lang === 'node') {
+    language.httpsnippet[2] = {
+      useObjectBody: true,
```
erunion, line 49: You want to move this into `src/supportedLanguages` instead? Since it's not using `oasUrl` or `oas` like [node-simple](https://www.npmjs.com/package/api) snippets are we can probably leave this option contained there.

erunion (approved):

packages/oas-to-snippet/src/index.js
Dashron: Oh totally, way better. Didn't notice that config list before. Getting a new pr shortly.

PR #1065, opened 2020-11-24
feat: flipping the order of data in the response schema component

erunion:
## 🧰 What's being changed?

This flips the order of data we're showing in the response schema component on operations from `type` → `property` to `property` → `type`.

## 🧪 Testing

You can test this out on the petstore example under the "Finds Pets by status" operation. Some examples:

Screen Shot 2020-11-24 at 10 51 56 AM

Screen Shot 2020-11-24 at 10 53 31 AM

### DIFF
--- a/packages/api-explorer/src/ResponseSchemaBody.jsx
+++ b/packages/api-explorer/src/ResponseSchemaBody.jsx
@@ -55,6 +55,8 @@ class ResponseSchemaBody extends React.Component {
               textAlign: 'right',
               overflow: 'visible',
               wordBreak: 'break-all',
+              // eslint-disable-next-line no-dupe-keys
+              wordBreak: 'break-word',
             }}
           >
             {row.name}
--- a/packages/api-explorer/src/ResponseSchemaBody.jsx
+++ b/packages/api-explorer/src/ResponseSchemaBody.jsx
@@ -51,24 +51,23 @@ class ResponseSchemaBody extends React.Component {
         <tr key={Math.random().toString(10)}>
           <th
             style={{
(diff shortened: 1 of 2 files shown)


packages/api-explorer/src/ResponseSchemaBody.jsx
```diff
@@ -51,24 +51,23 @@ class ResponseSchemaBody extends React.Component {
         <tr key={Math.random().toString(10)}>
           <th
             style={{
-              whiteSpace: 'nowrap',
-              width: '30%',
               paddingRight: '5px',
               textAlign: 'right',
-              overflow: 'hidden',
+              overflow: 'visible',
+              wordBreak: 'break-all',
```
runnabro, line 57: `break-word` will prefer to keep words intact while allowing breaks.

break-all: pricingOptions = pricingOption (break) s
Screen Shot 2020-11-24 at 12 06 08 PM

break-word: pricingOptions kept intact
Screen Shot 2020-11-24 at 12 06 13 PM

Doesn't work in IE, but we could have both and IE will take the break-all fallback.

```
wordBreak: 'break-all',
wordBreak: 'break-word',
```

runnabro (approved):
Left a small comment!

packages/api-explorer/src/ResponseSchemaBody.jsx
erunion: k i'll use both

erunion: @runnabro Updated!

PR #1070, opened 2020-11-25
fix(examples): fix examples button type

runnabro:
Before (Firefox and chrome only):
Screen Shot 2020-11-24 at 5 12 32 PM

After:
Screen Shot 2020-11-24 at 5 39 11 PM


erunion (approved):
lol I never noticed this

PR #1080, opened 2020-12-03
chore(deps): move variable to peerdep

kellyjosephprice:
## 🧰 What's being changed?

This sets `@readme/variable` as a peerDep. Variable is using a context to propagate user variables. So, it needs to be a singleton in the app, otherwise any imported consumers won't be matched the correct provider.

[ticket](https://app.asana.com/0/1199089258199423/1199528380076880/f)


Dashron:
code looks fine 👍  is this one of those "if the site loads it's fine" issues, or is there something else we should look for.

erunion:
@Dashron Yeah if the site loads and tests pass it's fine.

PR #1082, opened 2020-12-04
fix(api-explorer): fix legacy status ranges (4XX, etc)

Dashron:
## 🧰 What's being changed?

Before this PR, legacy apps that used status code ranges (1XX, 2XX, 3XX, 4XX and 5XX) in their responses failed to show those responses in the non-legacy API explorer.

## 🧪 Testing

- Create a legacy project by disabling the new api explorer in the project configuration
- Manually create an API call with a response status code of 4XX. Make sure to include an example response body.
- View the project in the legacy API explorer, and see the 4XX status code in the response examples
- Enable the new API explorer from the project configuration
- View the project in the new API explorer. Before this PR you would not see the 4XX status code. After this PR you should see it.


erunion (approved):
giphy

PR #1092, opened 2020-12-07
fix: only access the proxy-enabled extension if we need to

erunion:
## 🧰 What's being changed?

* [x] Changes `@readme/oas-to-har` to only look for the `x-proxy-enabled` OAS extension if the option to do so has been passed into the library.

### DIFF
--- a/packages/oas-to-har/src/index.js
+++ b/packages/oas-to-har/src/index.js
@@ -81,10 +81,10 @@ module.exports = (
     httpVersion: 'HTTP/1.1',
   };
 
-  const proxyEnabled = extensions.getExtension(extensions.PROXY_ENABLED, oas, operation);
-  // TODO look to move this to Oas class as well
-  if (proxyEnabled && opts.proxyUrl) {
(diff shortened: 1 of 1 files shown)


packages/oas-to-har/src/index.js
```diff
@@ -81,10 +81,10 @@ module.exports = (
     httpVersion: 'HTTP/1.1',
   };
 
-  const proxyEnabled = extensions.getExtension(extensions.PROXY_ENABLED, oas, operation);
-  // TODO look to move this to Oas class as well
-  if (proxyEnabled && opts.proxyUrl) {
-    har.url = `https://try.readme.io/${har.url}`;
+  if (opts.proxyUrl) {
```
gratcliff, line 84: Do we not need to check for `proxyEnabled` any longer?

erunion: `proxyEnabled` was the result of the `getExtension` call and is still in use on line 85.

Dashron (approved):
code looks good

PR #1095, opened 2020-12-09
refactor: use nanoid instead of shortid

darrenyong:
## 🧰 What's being changed?
| 🎫 [Asana Task](https://app.asana.com/0/1199087650851461/1199553290437389/f) |
|--------------------------------------------------------------------------------------|

[shortid has been deprecated](https://www.npmjs.com/package/shortid). They recommend using `nanoid` — this just swaps `shortid` with `nanoid`.

## 🧪 Testing
 Click on the Add button for those params to make sure you can add rows. 

1. Open up the local environment
2. Open a Swagger file that has object and array type parameters. `types.json` is a good one cause it will have all different types of params!
3. Find any parameter that requires you to Add a row of input like `array of strings` or `array of objects`. 
4. Click Add! If you can add rows, it should be good to go 😄

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22059,9 +22059,9 @@
       "dev": true
     },
     "nanoid": {
-      "version": "2.1.11",
-      "resolved": "https://registry.npmjs.org/nanoid/-/nanoid-2.1.11.tgz",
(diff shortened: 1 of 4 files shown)


Dashron:
Darren pointed me to http://localhost:9966/?selected=swagger-files%2Ftypes.json which seems to work great.

Dashron (approved):

erunion (approved):

PR #1096, opened 2020-12-09
fix: bug where it takes two clicks to get a code snippet to copy

erunion:
## 🧰 What's being changed?

This fixes a bug where it takes two clicks on the code sample copy button to copy the snippet you're looking at.

## 🧪 Testing

Bug is in production: https://preview.readme.io/

1. Click the copy button on a code sample. The code you're looking at should exist in your clipboard.
2. Click on another language tab.
3. Click the copy button. Code from the previous tab will still be in your clipboard.
4. Click the copy button again. Code from the new tab should be in your clipboard.

### DIFF
--- a/packages/api-explorer/__tests__/CopyCode.test.jsx
+++ b/packages/api-explorer/__tests__/CopyCode.test.jsx
@@ -1,0 +1,37 @@
+const React = require('react');
+const { mount } = require('enzyme');
+
+const CopyCode = require('../src/CopyCode');
+
+const curl = `curl --request POST \
+--url http://petstore.swagger.io/v2/pet \
+--header 'Authorization: Bearer 123' \
+--header 'Content-Type: application/json'`;
+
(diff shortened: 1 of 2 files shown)


domharrington (approved):
LGTM. Any idea how this could've happened? Or how long this has been an issue for?

erunion:
I'm not sure how long this has been in place. I pushed some minor proptype updates to `CopyCode` with the JSON editor to fix some test warnings, but reverting those locally it didn't fix it. The `CopyCode` usage in `CodeSample` also hasn't changed in 2 years.

Might've been a side effect of some React updates we did with the JSON editor?

erunion:
Ticket that spawned this was created back on October 22nd which is the day after the initial tutorials work was integrated in in https://github.com/readmeio/api-explorer/pull/958 so it's possible React issues with that did it.

PR #1110, opened 2020-12-15
feat: oas dereferencing and performance improvements

erunion:
_This is rewrite of https://github.com/readmeio/api-explorer/pull/1097._

## 🧰 What's being changed?

* [x] Bubbles up the OAS dereferencing work we're doing for response examples up into a new `DocAsync` component that wraps `Doc` so OAS definitions are dereferenced before the doc and operations load.
* [x] `ResponseExample`, `ResponseTabs`, and `ResponseSchemaBody` components are no longer async as they're now supplied a fully dereferenced OAS.
* [x] Minor performance improvements to the `ResponseExample` component for APIs from our manual editor. Previously we were upgrading their responses into a shape that we could handle within the `render` method, causing that work to happen anytime the component changed. Since this work should be static it's now happening within the constructor of the `Response` component.

This also includes some incredible performance improvements to how we handle OAS definitions and the `oas` tooling library:

* [x] The root API Explorer component now only creates a single `oas/tooling` instance for each OAS definition we're handling.
* [x] If `await oas.dereference()` is called twice and the second call happens before the first finishes, the second call will now wait for the first to finish -- the second won't also dereference.
  * [x] Additionally if `oas.dereference()` will now only ever run dereference processing once irregardless of when or how many times it was called.
* [x] Request and response example schema discovery and generation is now memoized (with [memoizee](https://www.npmjs.com/package/memoizee) to reduce the load of compiling the same common schemas over and over.

## 🧪 Testing

https://readme-api-e-feat-deref-hy7jrl.herokuapp.com/

Since the crux of this work is to create a wrapper on the `Doc` component to do async work, we need to verify the following:

* [ ] Dereferencing actually works and API definitions with `$ref` pointers get resolved properly.
* [ ] Resolved `$ref` pointers show up in the form as expected.
* [ ] Sample responses are still being populated.
  * https://readme-api-e-feat-deref-hy7jrl.herokuapp.com/?selected=swagger-files%2Fexamples.json
* [ ] Our existing circular ref handling is unchanged. 
  * https://readme-api-e-feat-deref-hy7jrl.herokuapp.com/?selected=swagger-files%2Fcyclical-refs.json

## 🐎  Performance Improvements

We have a couple main performance issues right now within the API Explorer:

1. When we generate sample requests and responses we dereference the API definition. The issue with this though is that that dereferencing work happens for each individual operation within the definition so if an API has hundreds of operations, and each of those are fairly `$ref` heavy we'll be doing a **lot** of work on page load.
    * Related to this, because we're generating individual `oas` objects for each operation, but they're using the same API definition variable in memory this sometimes runs the chance of corrupting the API definition and `$ref` pointers getting lost as it's possible the API definition gets dereferenced multiple times at simultaneously.
2. Not as heavy as an issue as the above, but when we do generate sample requests and responses, we often re-run the same conversion code for the same primitives. For example if a response has 20 strings, we'll be converting 20 strings to `string` 20 times for that response. If those 20 strings are in a `$ref` we'll be converting those 20 strings *n* times for however many times its used. The code for handling this conversion is pretty slim, but it's still a lot of unnecessarily duplicative work.
3. Since we also create an `oas` object for each operation, we also blow up the memory usage of the API Explorer on large definitions as it's duplicating the same large spec multiple times.

To resolve these issues we're now doing a few things:

* Creating a single `oas` object for every API definition. `oas` objects are then shared across each operation.
* Implementing promise waiting on OAS dereferencing so it'll only ever happen once.
* Memoizing request and response sample generation.

The results of this work are quite remarkable with one particularly heavy API definition going from 17 seconds to load down to 5.

https://user-images.githubusercontent.com/33762/103586803-b4ca0280-4e9a-11eb-8570-b59e3dd7715e.mp4

### Why do this?
A couple reasons!

1. I am currently working on exposing request examples in the form by way of input placeholders and in order to do that I need to merge our [getParametersAsJsonSchema](https://github.com/readmeio/oas/blob/master/tooling/operation/get-parameters-as-json-schema.js) library with [sampleFromSchema](https://github.com/readmeio/oas/blob/master/tooling/samples/index.js). Problem is that the latter expects a dereferenced  definition and the former does `$ref` lookups.
2. Circular schemas. With the [json-schema-ref-parser](https://apitools.dev/json-schema-ref-parser/) library that we're using, and have been using for the past month to generate response examples, we now have slightly better handling for circular schemas in that we don't fully crap out the explorer when we encounter them with a maximum call stack error. `json-schema-ref-parser` has some options to ignore circular schemas and leave the `$ref` intact within the OAS so this allows us to quickly detect a circular schema so we can ignore it without having to write our own recursive traversal mapping logic.
2. Performance reasons. Since we currently don't memoize `$ref` lookups, we end up doing a lot of the same work multiple times depending on how often a `$ref` is used within an OAS. By dereferencing the OAS up front we now no longer have `$ref` components anywhere within the spec and don't need to fetch schemas on demand.
    * There's still a lot of work to be done here with regard to memoizing as large APIs are still extremely slow to render due to the response example generation. I'm working on more improvements for this in separate PRs.

### Why dereference here?
Instead of doing this dereference work within the explorer and creating these async components, why not send the explorer a fully dereferenced OAS instead? Good question! The problem with doing that is right now for our legacy Swagger definitions we're storing them in the DB with [circular-json](https://www.npmjs.com/package/circular-json) in order for us to be stringify circular schemas. We currently do not do this for OAS definitions.

It's currently an open question whether we still need to do this, or if we could stringify a dereferenced (while ignoring circular schemas as we do in `oas`, see [this code](https://github.com/readmeio/oas/blob/master/tooling/index.js#L222-L247)), but until we solve that this is the way to go.


erunion:
Getting `anyOf` issues on https://readme-api-e-feat-deref-hy7jrl.herokuapp.com/?selected=swagger-files%2Ftypes.json, looking into it.

packages/api-explorer/__tests__/Doc.test.jsx
```diff
@@ -510,15 +524,13 @@ describe('#enableRequestBodyJsonEditor', () => {
...
+          oas={oas}
         />
       );
 
-      return waitFor(() => {
-        expect(doc.state('formDataJson')).toStrictEqual(petExample);
-        expect(doc.state('formDataJsonOriginal')).toStrictEqual(doc.state('formDataJson'));
-        expect(doc.state('formDataJsonRaw')).toMatch('"status": "available"');
-      });
+      expect(doc.state('formDataJson')).toStrictEqual(petExample);
```
domharrington, line 531: Much nicer.

packages/api-explorer/src/Doc.jsx
```diff
@@ -70,69 +70,56 @@ class Doc extends React.Component {
...
 
-    this.operation
-      .getRequestBodyExamples()
-      .then(examples => {
-        if (!Object.keys(examples).length) {
-          return;
-        }
+    // If we need to build out the request body JSON editor let's pull examples to fill it with.
+    if (this.shouldEnableRequestBodyJsonEditor()) {
+      const examples = this.operation.getRequestBodyExamples();
```
domharrington, line 88: How did this become synchronous? Where's that code?

packages/api-explorer/src/Doc.jsx
```diff
@@ -70,69 +70,56 @@ class Doc extends React.Component {
...
-            formDataJson: code,
-            formDataJsonOriginal: code,
-            formDataJsonRaw: stringifyPretty(code),
-          });
-        }
-      })
-      .catch(() => {
-        // If we fail to generate examples for whatever reason fail silently.
-      });
+        this.state.formDataJson = code;
```
domharrington, line 118: So we can attach it straight onto the state now cos it's happening synchronously?

packages/api-explorer/src/DocAsync.jsx
```diff
@@ -0,0 +1,55 @@
...
+
+const Doc = require('./Doc');
+const ErrorBoundary = require('./ErrorBoundary');
+
+const DocAsync = props => {
+  const { maskErrorMessages, oas, onError, ...docProps } = props;
+  const [loading, setLoading] = useState(true);
+  const [error, setError] = useState();
+
+  useEffect(() => {
```
domharrington, line 13: Woah this is some crazy code haha

packages/api-explorer/src/DocAsync.jsx
```diff
@@ -0,0 +1,55 @@
...
+    // rendering out that error message.
+    return (
+      <ErrorBoundary appContext="explorer" error={error} maskErrorMessages={maskErrorMessages} onError={onError}>
+        {error.message}
+      </ErrorBoundary>
+    );
+  }
+
+  if (loading) {
+    return null;
```
domharrington, line 37: Maybe this is where we can render a stubbed version of the <Doc> to improve perceived performance in future?

domharrington (approved):
The code for this looks sound! Lots of simplified components due to the top level dereferencing. Pretty happy with this but it is quite scary due to the amount of change. Any idea on next steps for this?

packages/api-explorer/src/Doc.jsx
erunion: @domharrington That happened in https://github.com/readmeio/oas/pull/327

packages/api-explorer/src/Doc.jsx
erunion: Yep, it's no longer a promise so we don't need to do all this crazy handling.

packages/api-explorer/src/DocAsync.jsx
erunion: 😬

packages/api-explorer/src/DocAsync.jsx
erunion: Yeah a loading spinner, or maybe something like Facebook does where they load in opaque boxes in that get swapped out as components load.

erunion:
@domharrington I'd like to get a build going of it later this week and put it up on a CI app to have people run through on some client sites to see where/if it breaks down.

PR #1114, opened 2020-12-17
fix: making the json editor only be present on body parameters

erunion:
## 🧰 What's being changed?

* [x] Fixes a bug in the new JSON request body editor where it'd appear for non-request bodies if the operation had more than just that for parameters (like a path param).

## 🧪 Testing

You can see this buggy behavior on https://preview.readme.io/ within of the "Updated user" operation:

Screen Shot 2020-12-17 at 9 10 49 PM

Clicking on the JSON tab would make it look like both tabs were engaged because both tabs for both the path and body sections were being placed on top of each other:

Screen Shot 2020-12-17 at 9 12 15 PM

What it should look like:

Screen Shot 2020-12-17 at 9 11 11 PM

And with the tab selected, the double highlight bug is no longer present because there are no longer multiple tab components on the page:

Screen Shot 2020-12-17 at 9 13 10 PM


packages/api-explorer/src/Params.jsx
```diff
@@ -128,8 +128,8 @@ class Params extends React.Component {
           this.jsonSchema.map(schema => {
             return (
               <React.Fragment key={`${schema.type}-block`}>
-                {enableJsonEditor ? (
-                  <React.Fragment>
+                {enableJsonEditor && schema.type === 'body' ? (
+                  <div className="param-type-body">
```
runnabro, line 132: This change here is a small risk to breaking any custom CSS if a customer has a specific selector or prop that only affects direct child node (ie. `.param-type -> .param-type-header`, or `.param-type {display: flex}`)

erunion: Should be fine since we just have Envoy using this stuff right now

erunion: Oh I see. Will test!

erunion: @runnabro Keeping both `position: relative` rules around it seems okay?

Screen Shot 2020-12-17 at 8 53 39 PM

runnabro: Oh, we don't actually need both, ill remove the parent one, thanks!

PR #1124, opened 2020-12-22
test: building out a framework for testing readme manual apis

erunion:
## 🧰 What's being changed?

* [x] Builds out a framework for us to be able to better test ReadMe manual APIs without having to link up the API Explorer with the ReadMe codebase.

This is done by dropping a specification into `example/swagger-files` that contains a the following properties: 
  * `readmeManual`
  * `appearance` *(optional)*
  * `flags` *(optional)*
  * `docs`
  * `oasFiles` *(optional)*
  * `oasUrls` *(optional)*

The structure of this data is identical to how we place it into ReadMe projects with the `readme-data-oasFiles` hidden inputs.

## 🧪 Testing

I've placed a test sample project from Logitech Circle in here that replicates their [Notification API](https://developers.logitech.com/circle/reference/notification-events) docs with all internal IDs swapped out with random UUIDs.

http://localhost:9966/?selected=swagger-files%2Fmanual-api-logitech.json


erunion:
@rafegoldberg @kellyjosephprice There's currently a bug with this where content is being duplicated:

Screen Shot 2020-12-21 at 4 48 00 PM

I'm not sure what setting is controlling this, but it's not happening in production. Any ideas what's going on here?

kellyjosephprice:
> I'm not sure what setting is controlling this, but it's not happening in production. Any ideas what's going on here?

Here's what I know so far. The logitech production page is using the old markdown engine. It looks like the old markdown engine returns nothing when it's passed the `isThreeColumn="right"` (I didn't dig into the library, but that looks like the behavior). It seems like the new `markdown` library should probably do the same?

rafegoldberg:
@erunion @kellyjosephprice I actually think this is [a markup issue](https://github.com/readmeio/api-explorer/blob/cd1daafa1cfbbfc62ff112b2dc495744d17476bc/packages%2Fapi-explorer%2Fstyle%2Fmarkdown-overrides.scss#L21-L22) with how we manage pinned markdown content. Try firing up the dev server on this branch and adding an `id=api-explorer` attribute to the `<body>` element... Fixes this issue for me!

rafegoldberg:
This works in-app because [we add that ID to the React root in our ReadMe jade](https://github.com/readmeio/readme/blob/e8d02db5ae61ebadd5539b28399f96c2e91c96b8/packages%2Fhub%2Fviews%2Freference%2Freference.jade#L4)!

https://github.com/readmeio/readme/blob/e8d02db5ae61ebadd5539b28399f96c2e91c96b8/packages%2Fhub%2Fviews%2Freference%2Freference.jade#L4

kellyjosephprice:
> This works in-app because [we add that ID to the React root in our ReadMe jade](https://github.com/readmeio/readme/blob/e8d02db5ae61ebadd5539b28399f96c2e91c96b8/packages%2Fhub%2Fviews%2Freference%2Freference.jade#L4)!

That makes sense.

kellyjosephprice:
That is too easy to do.

Screenshot from 2020-12-23 10-35-20

erunion:
Oh rad, Happy Life Day!

giphy

erunion:
That fixed it! Also makes the Markdown content look a lot better too.

Screen Shot 2020-12-23 at 11 22 31 AM

Dashron (approved):

PR #1128, opened 2021-01-05
feat: add support for parameter serialization

Dashron:
## 🧰 What's being changed?

This PR adds support for all [style](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.1.0.md#style-values) values on all relevant oas parameters. Style is OAS's way of supporting URI templates.

## 🧪 Testing

Load up the API Explorer example and play with the following files
- `parameter-styles-cookie.json`
- `parameter-styles-headers.json`
- `parameter-styles-path.json`
- `parameter-styles-query.json`

Compare them to the examples listed here: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.1.0.md#style-examples and here: https://swagger.io/docs/specification/serialization

The elements I show in those OAS files are the only elements considered "valid" by the spec. The tests though cover every valid and invalid input type to ensure we have complete adherence to the spec.

**Why the 3.1.0 spec?** 3.1.0 cleans up confusion in the previous specs by ensuring the Style Values and Style Examples tables are consistent with each other, and with the URI template spec. I believe everything described in 3.1.0 is supposed to work in previous versions.

## Additional Notes

There are two features in the spec that this PR does not currently support, because swagger's encoding library doesn't support them. If important enough, I can figure out how to support them. I have listed the cases below, and included commented out tests in this PR.
- spaceDelimited style, explode=false, object input.
- pipeDelimited style, explode=false, object input.

Also, when testing you will notice that commas in the cookie headers are encoded. It's hard to tell if this is intentional or not due to the fact that the cookie spec differs from all other header specs. Because this encoding happens in httpsnippet and not this oas code, I have left it as is for the moment (see https://github.com/readmeio/httpsnippet/blob/master/src/index.js#L94).


packages/oas-to-har/src/index.js
```diff
@@ -6,8 +6,15 @@ const parseDataUrl = require('parse-data-url');
 
 const configureSecurity = require('./lib/configure-security');
 const removeUndefinedObjects = require('./lib/remove-undefined-objects');
+const formatStyle = require('./lib/style-formatting');
 
 function formatter(values, param, type, onlyIfExists) {
+  if (param.style) {
+    // Note: Technically we could send everything through the format style and choose the proper default for each
```
Dashron, line 13: This is a comment for potential future cleanup. Technically non-styled formats are coming through with a default style, so we could push all parameters through this style logic instead of branching.

packages/oas-to-har/__tests__/lib/style-formatting/index.test.js
```diff
@@ -0,0 +1,948 @@
...
+      ],
+      [
+        'should NOT support space delimited query styles for exploded array input',
+        paramExplode,
+        { query: { color: arrayInput } },
+        [],
+      ],
+      // This is supposed to be supported, but the style-seralizer library we use does not have support. Holding off for now.
+      /* [
+        'should support space delimited query styles for non exploded object input',
```
Dashron, line 534: Here's one of the unsupported tests I mention in the PR description

example/swagger-files/parameter-styles-cookie.json
```diff
@@ -0,0 +1,131 @@
...
+      "license": {
+        "name": "Apache 2.0",
+        "url": "http://www.apache.org/licenses/LICENSE-2.0.html"
+      }
+    },
+    "paths": {
+        "/nonExplodedFormCookies": {
+            "get": {
+                "summary": "Demonstrate all non-exploded form cookie styles",
+                "operationId": "demonstrateCookies",
```
erunion, line 19: Can you make the operationId's in these definitions? There's some DOM warnings being thrown because `id` attributes are getting duplicated.

Screen Shot 2021-01-06 at 10 26 59 AM

example/swagger-files/parameter-styles-cookie.json
```diff
@@ -0,0 +1,131 @@
...
+                "responses": {
+                    "200": {
+                        "description": "Successful response"
+                    }
+                }
+            }
+        },
+        "/explodedFormCookies": {
+            "get": {
+                "summary": "Demonstrate all exploded form cookie styles",
```
erunion, line 71: Is it right that properties within `object` get hoisted up when exploded?

Screen Shot 2021-01-06 at 10 31 34 AM

example/swagger-files/parameter-styles-query.json
```diff
@@ -0,0 +1,247 @@
...
+                    "schema": {
+                        "type": "array",
+                        "items": {
+                            "type": "string"
+                        }
+                    }
+                }, {
+                    "name": "object",
+                    "in": "query",
+                    "description": "spaceDelimited style object. Note: This does not currently work. See associated test.",
```
erunion, line 142: Haha I was like "hmm this isn't working" and then noticed this comment.

Screen Shot 2021-01-06 at 10 42 56 AM

example/swagger-files/parameter-styles-query.json
```diff
@@ -0,0 +1,247 @@
...
+                "responses": {
+                    "200": {
+                        "description": "Successful response"
+                    }
+                }
+            }
+        },
+        "/explodedDeepObjectQueries": {
+            "get": {
+                "summary": "Demonstrate all non-exploded deep object query styles",
```
erunion, line 210: This one's throwing some errors right now on a fresh page load.

Screen Shot 2021-01-06 at 10 44 35 AM

packages/oas-to-har/__tests__/lib/style-formatting/index.test.js
```diff
@@ -0,0 +1,948 @@
...
+const oasToHar = require('../../../src/index');
+
+const oas = new Oas();
+
+const emptyInput = '';
+const undefinedInput = undefined;
+const stringInput = 'blue';
+const arrayInput = ['blue', 'black', 'brown'];
+const undefinedArrayInput = [undefined];
+const objectInput = { R: 100, G: 200, B: 150 };
```
erunion, line 12: I'm afraid to ask but how does it handle nested objects, arrays of objects, or arrays of arrays?

packages/oas-to-har/src/lib/style-formatting/index.js
```diff
@@ -0,0 +1,117 @@
...
+
+// Certain styles don't support empty values, This function tracks that list
+function shouldNotStyleEmptyValues(parameter) {
+  return ['simple', 'spaceDelimited', 'pipeDelimited', 'deepObject'].includes(parameter.style);
+}
+
+// Note: This isn't necessarily part of the spec. Behavior for the value 'undefined' is, well, undefined.
+//   This code makes our system look better. If we wanted to be more accurate, we might want to remove this,
+//   restore the un-fixed behavior for undefined and have our UI pass in empty string instead of undefined.
+function removeUndefinedForPath(value) {
```
erunion, line 11: Can you use https://github.com/readmeio/api-explorer/blob/next/packages/oas-to-har/src/lib/remove-undefined-objects.js?

erunion (changes requested):
Incredible work.

example/swagger-files/parameter-styles-cookie.json
Dashron: Correct, that's basically the definition of explode. Turn each key=>value pair into a top level key. There are some edge cases that don't respect this though, and I've called it out in the tests.

example/swagger-files/parameter-styles-query.json
Dashron: Yeah that's one of the ones called out in the additional notes. Seems to be a problem with the encoding library we use. I can easily fix it, but hesitated.

packages/oas-to-har/src/lib/style-formatting/style-serializer.js
```diff
@@ -0,0 +1,202 @@
+/* eslint-disable sonarjs/no-extra-arguments, no-use-before-define, sonarjs/prefer-immediate-return, no-param-reassign, sonarjs/no-identical-functions */
+
+/**
+ * This file has been extracted and modified from Swagger UI.
+ *
+ * @license Apache 2.0
+ * @link https://github.com/swagger-api/swagger-js/blob/e5d83c7c8395baf1e30a16e636aabe3bba21360e/src/execute/oas3/style-serializer.js
+ */
+
+const { Buffer } = require('buffer');
```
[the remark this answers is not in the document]
Dashron: This is the file I pulled from the swagger-js. Happy to update it if you think it's worth it.

packages/oas-to-har/package.json
```diff
@@ -18,6 +18,7 @@
   },
   "dependencies": {
     "@readme/oas-extensions": "^10.0.8",
+    "lodash": "^4.17.19",
```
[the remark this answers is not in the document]
Dashron: Oh yeah, that was probably part of the style-serializer code I ripped out. I'll pull it.

packages/oas-to-har/src/lib/style-formatting/style-serializer.js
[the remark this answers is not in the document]
erunion: Can you do a build size comparison with and without it? It's working as is so if it doesn't impact the build size at all might as well leave it.

You can do `npm run build` compile the dists and see.

packages/oas-to-har/__tests__/lib/style-formatting/index.test.js
Dashron: I'm having a hard time finding it, but somewhere said that nested objects or arrays are not allowed. I think it might be somewhere in the URI template spec that each style is from. Still looking

As for what would happen, I believe it would apply javascript stringification (via `String(val)`), so we would get something like `R=[object]`

erunion: Works for me. I can't imagine anyone actually trying to use nested data in a URI template anyways.

example/swagger-files/parameter-styles-cookie.json
Dashron: fixed locally

example/swagger-files/parameter-styles-query.json
Dashron: fixed locally

packages/oas-to-har/src/lib/style-formatting/index.js
Dashron: Spent some time testing this, and I can't because the purpose is not to remove undefined, but to make undefined empty strings.

packages/oas-to-har/src/index.js
Dashron: Just a note for future reference, the spec actually very clearly defines what all those defaults are.

packages/oas-to-har/src/lib/style-formatting/style-serializer.js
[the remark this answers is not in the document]
Dashron: spoke with Jon, I don't see a file size improvement with or without it, so I'm going to keep it.

Dashron:
@erunion  all your requested changes should be up

erunion (approved):

packages/oas-to-har/__tests__/lib/style-formatting/index.test.js
```diff
@@ -0,0 +1,948 @@
...
+ *  Style Examples
+ *
+ *  Assume a parameter named color has one of the following values:
+ *
+ *     string -> "blue"
+ *     array -> ["blue","black","brown"]
+ *     object -> { "R": 100, "G": 200, "B": 150 }
+ *
+ *  The following table shows examples of rendering differences for each value. (NON-SPEC-NOTE: I've added the additional `in` column for easy reference)
+ *  style 	        explode 	empty 	string 	      array 	                              object                                  in
```
domharrington, line 52: Can we tidy up the whitespace on this comment so we can see the alignment a little clearer?

packages/oas-to-har/__tests__/lib/style-formatting/index.test.js
```diff
@@ -0,0 +1,948 @@
...
+        'should support label path styles styles for non exploded array input',
+        paramNoExplode,
+        { path: { color: arrayInput } },
+        'https://example.com/style-path/.blue.black.brown',
+      ],
+      [
+        'should support label path styles styles for exploded array input',
+        paramExplode,
+        { path: { color: arrayInput } },
+        'https://example.com/style-path/.blue.black.brown',
```
domharrington, line 251: I have absolutely no idea why anyone would ever want to do this 🤷‍♂️

domharrington (approved):
This is perhaps one of the most comprehensively tested PR's I've ever seen. Code looks great, and I'm sure you now know more than you ever thought you would regarding OAS Style values. 

One potential improvement would be to move the style-formatting to a separate npm module, but that's so minor and It definitely doesnt need to be done now.

Is there anything else that needs to be done here? Or is this completely backwards compatible?

packages/oas-to-har/__tests__/lib/style-formatting/index.test.js
Dashron: I would hope anyone doing anything like this uses query params instead of the path. It's definitely bizarre.

packages/oas-to-har/__tests__/lib/style-formatting/index.test.js
Dashron: looks like there's a weird mix of spaces and tabs. fixing.

Dashron:
It's completely backwards compatible, so we just need to send it out. I'm still on the fence about cookies though. I think the encoding we do here is incorrect, but it will be hard to get verification. In either circumstance we wouldn't fix it in this repo, that's in httpsnippet.

If cookies are wrong, people who use the cookie style and follow the example request will send data that is not properly parsed by the receiving server. Instead of the server receiving a list of cookies to interpret, they would get one cookie that had commas in the value.

As for the tests, it was essential to get everything right. This was my most thorough application of test driven development, as the tests were done far before the code was and helped me find all the weird edge cases. There's actually more tests I thought might be useful, but I cut it off at "just the examples table". This is such a complex feature and the documentation is good enough just until you need clarification, and then it's useless. 

The comment table fix is pushed up and ready to go.

PR #1146, opened 2021-01-14
fix(api-explorer): prevent nulls from breaking operations

Dashron:
## 🧰 What's being changed?

Null values or defaults had the potential of throwing an error when rendering a non-body operation. This fixes that.

## 🧪 Testing

Message me for an example json file and put it in your `examples/swagger-files` folder, run `npm start`, and test the `manual-api-keeptrucking.json` file. Look for the endpoint named `driver_utilization`. Before this change you would see an error. After it you should see a rendered operation.

### DIFF
--- a/packages/oas-to-har/__tests__/index.test.js
+++ b/packages/oas-to-har/__tests__/index.test.js
@@ -182,6 +182,49 @@ describe('parameters', () => {
         { query: { id: 0 } },
         [{ name: 'id', value: '0' }],
       ],
+      [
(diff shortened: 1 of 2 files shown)


domharrington (approved):
LGTM. What was the commit/pr that broke it?

erunion (approved):

Dashron:
@domharrington  this break was introduced as part of the styles PR (https://github.com/readmeio/api-explorer/pull/1128)

PR #1147, opened 2021-01-15
fix(api-explorer): stop x-readme-api-explorer header from breaking unproxied cors requests

Dashron:
## 🧰 What's being changed?

`x-readme-api-explorer` header is causing issues for some customers due to cors: https://app.intercom.com/a/apps/m6855w1q/inbox/inbox/conversation/7278100009838. We could just remove this for people not using the proxy, but it sounds like we're not using this header at all, so we chose to remove it entirely (https://readmeio.slack.com/archives/C5B798RFZ/p1610669718031200)

## 🧪 Testing

Test any API request from the explorer with the network tab open, and you shouldn't see the `x-readme-api-explorer` header being sent.

### DIFF
--- a/packages/api-explorer/src/Doc.jsx
+++ b/packages/api-explorer/src/Doc.jsx
@@ -25,8 +25,6 @@ const { Operation } = Oas;
 const parseResponse = require('./lib/parse-response');
 const Content = require('./block-types/Content');
 
-const { version: packageVersion } = require('../package.json');
-
 const stringifyPretty = data => JSON.stringify(data, undefined, 2);
 
 class Doc extends React.Component {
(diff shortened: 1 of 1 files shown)


domharrington (approved):

erunion (approved):

bridget-basiq:
Hi guys, I am still seeing this being returned when using the explorer without the proxy. Was this reverted or is it possible I'm using an older/incorrect version of README? :) @domharrington @Dashron

erunion:
@bridget-basiq Are you seeing this on a ReadMe project, or are you running this project yourself? This project is no longer being used on new ReadMe projects, but a variant of it that we have is still sending this header. What's the problem it's causing you?

bridget-basiq:
Thanks Jon! I am using readme directly, not running myself - so perhaps I’m
using an older version? I’m getting the same error described.
X-readme-api-explorer header not allowed when using explorer without proxy.

Is there a way to check/change the version We are on?

heythisisnate:
Having a problem with this as well. The paid/hosted Readme.com is sending this header in API requests using the Try It feature, which is failing due to CORS header validation, causing the API call to fail.

ashmanmalik:
Still having issues and got this error: 
X-readme-api-explorer header not allowed when using explorer without proxy.
Any way to fix this?

erunion:
A fix for this will be going out in our daily deploy tomorrow morning. I'm terribly sorry for all the trouble this has caused yall for so long.

PR #1149, opened 2021-01-20
perf: oas dereferencing when we load operations, not docs

erunion:
## 🧰 What's being changed?

So with the recent OAS dereferencing and perf work in #1110 we've introduced blocking on the `Doc` component until OAS files are fully dereferenced. This works plenty fine for customers who have a single OAS file, but unfortunately for some of our customers we've told them to chunk up their OAS files into smaller files to boost perf. Problem with this is that we're now also dereferencing those OAS files when we load the page.

For a customer that has 15 OAS files on their project split up across a couple dozen doc pages, we're now dereferencing all 15 of those OAS files before we load even a single line of documentation. You can see this in action with this example:

https://user-images.githubusercontent.com/33762/105113779-775c9c00-5a7a-11eb-9ad9-1dea638c83d0.mp4

Yikes!

To help with this I've slightly refactored our `DocAsync` component to be `EndpointAsync` and moved it into where we do `renderEndpoint` within the `Doc`. This takes advantage of our existing `Waypoint` handling so when a doc is in view we'll now dereference the OAS that it uses, versus dereferencing that all upfront. And it works! Kinda...

Now docs are loaded almost immediately, but the rest of the content from the OAS files in a bit later after the dereference work happens. It's faster for sure, but it looks *weird*.

https://user-images.githubusercontent.com/33762/105114058-fd78e280-5a7a-11eb-9a9f-eeac978b5b16.mp4

## 🧪 Testing

To properly change this you need to have a full dump of an intense ReadMe project and all the data that we feed into the root `ApiExplorer` component for a page to load. DM me and I can get you a file to test with for one of these customers with many OAS files.


erunion:
@runnabro @domharrington @Dashron @mjcuva I'm currently pretty iffy on this at the moment. It's a huge improvement, but it just looks bad. Any ideas on how to make this look and feel good?

runnabro:
Couple options:
1. If the total load time is faster, we could hide all the content and show once everything is loaded. This would look how it was before, except faster.
2. Hide all the text/UI content, but show the UI that doesn't need to be loaded (similar to how Slack loads)

erunion:
@runnabro @Dashron Alright, I might be out of my element here. Things I've tried:

* Invoking it so we only dereference the OAS definition for the docs we're looking at.
  * Pretty much impossible to do with the way everything currently is because we load up a collection of `Doc` components immediately when the page loads, and if the user is came in with an anchor tag we can't send them directly there without parsing everything before it.
* Wrapping the `Doc` component with [react-waypoint](https://www.npmjs.com/package/react-waypoint) so we only initialize dereferencing when the user reaches the point in a page where the doc is.
  * Also failed for reasons similar to the first in this list as the anchor tag isn't present until the doc is there in the page, and if we waypoint it the anchor isn't created until the Doc is. This also spawned some really bizarre parallel loading issues where one OAS file would get loaded for dereferencing, then the second, then the 3/4/5/etc. Then they'd all finish at the same time. It ended up blowing up the page load for this Shipwell dump to a minute+.

Like @Dashron said I think we can possibly render the HTTP method and URL since that's immediately available without dereferencing, but what if we did some sort of shimmering effect on parameters, code samples, and responses while they're still being processed? Like this: https://nh9x1.csb.app/

runnabro:
@erunion Would the other endpoints and section titles still load before tho? That would cause the page to jump around a bit and I was trying to avoid that. I'm down to try it and see how it feels…I think this is one of those things we have to try and get a feel for it. Down to help too, just lmk!

erunion:
@runnabro TBH I'm not sure.

erunion:
We got a better solution going in https://github.com/readmeio/api-explorer/pull/1156

PR #1150, opened 2021-01-22
Perf/oas dereferencing loading

runnabro:
Adds loading state:
* Hides faster loading header content and borders
* Adds spinner to bottom-left

 Todo:
* Unhide header content and border when loading is complete
* Remove spinner when loading is complete

PR #1156, opened 2021-01-27
feat: introducing an interim loading screen in DocAsync

erunion:
| [☁️ &nbsp; CI App][demo] |
| --- |

## 🧰 What's being changed?

Pulling in some @runnabro CSS wizardy from https://github.com/readmeio/api-explorer/pull/1150 to make OAS dereferencing load a little smoother.

https://user-images.githubusercontent.com/33762/106051637-8e644500-609d-11eb-89c1-9067cd0fbbf7.mov

[demo]: https://readme-api-e-perf-deref-sxg7nh.herokuapp.com/


Dashron (approved):

packages/oas-form/src/utils.js
```diff
@@ -678,7 +678,7 @@ export function retrieveSchema(schema, rootSchema = {}, formData = {}) {
         }
       );
     } catch (e) {
-      console.warn(`could not merge subschemas in allOf:\n${e}`);
+      // console.warn(`could not merge subschemas in allOf:\n${e}`);
```
Dashron, line 681: not only weird that a console log made it through, but that we were testing it, thus expecting it.

erunion: RJSF has a number of `console.warn`'s around that they're also asserting are getting thrown.

PR #1157, opened 2021-01-27
Add support for discriminators

Dashron:
| [☁️ &nbsp; CI App][demo] |
| --- |

## 🧰 What's being changed?

Discriminators are fields in OAS to help manage oneOf and anyOf schemas. If a discriminator is defined, it indicates that a specific field within the oneOf or anyOf schema indicates which schema you should anticipate.

e.g. a "pet_type" field in cat and dog might be the discriminator for that schema. The value being "cat" or "dog". 

Discriminators also support mappings, where instead of the name of the schema, you can use a different value. Note: the field Title within a schema is not valid values, it must be the key of the schema within the components/schema object.

## 🧬 Testing

Load up production api explorer (https://preview.readme.io/) with either the pet_store, polymorphism, or discriminator examples, and compare it to this branch.

In the old setup you will notice an unlabeled dropdown, and a field that is common between all of the oneOf or anyOf values.

In the new setup you will notice that dropdown has a label, and that label happens to match the label of the field common between all of the oneOf or anyOf values. That field has also been removed from the oneOf or anyOf schemas, since it's been moved up a level.

The field should still appear in the examples, and changing the dropdown should change the selected schema..


packages/api-explorer/src/form-components/MultiSchemaField.jsx
```diff
@@ -0,0 +1,118 @@
+require('../../style/main.scss');
```
erunion, line 1: Shouldn't need to load this here fyi

packages/api-explorer/src/form-components/MultiSchemaField.jsx
```diff
@@ -0,0 +1,118 @@
...
+function extractDiscriminatorField(schema, property) {
+  if (schema.properties?.[property]) {
+    const discSchema = schema.properties[property];
+    delete schema.properties[property];
+    return discSchema;
+  }
+
+  // Sometimes discriminators are within each of the oneOf schemas, and not top level.
+  if (schema.oneOf) {
+    return extractDiscriminatorField(schema.oneOf[0], property);
```
erunion, line 18: It possible for the discriminator to be inside just one of the entries in a multi-array `oneOf`?

packages/api-explorer/src/form-components/MultiSchemaField.jsx
Dashron: cool, just copied it from another file, will remove

packages/api-explorer/src/form-components/MultiSchemaField.jsx
Dashron: When the discriminator is outside of the `oneOf`, it means that property MUST exist within the response payload, i.e. in each of the `oneOf`s. The behavior of discriminators within each of the oneOf schemas isn't clearly defined or regulated, but I would assume it works similarly. I haven't found any examples otherwise.

erunion: 👍  Doesn't look like the JSON Schema that the OAI offers validates that, but is probably a safe assumption for it to be in each. https://github.com/OAI/OpenAPI-Specification/blob/master/schemas/v3.0/schema.json

Dashron:
Notes to self before the weekend (and anyone who cares):
- This relies on a custom `MultiSchemaField`, where if there's a discriminator it uses a fork of the existing `MultiSchemaField` that uses a full `SchemaField` for the discriminator instead of just using a select widget. This allows us to use the custom template with the field name/label. If there's no discriminator it renders the standard `MultiSchemaField`
- What works: The discriminator field looks good, has a label, a drop down, and changing the dropdown changes the shown schema, and filling values in the form is reflected in the UI
- What I still need to do: the discriminator empties out on change, and isn't reflected in the examples, add support for mapping, ensure default values work, make sure everything works as if we were using the standard MultiSchemaField.
- Also: write tests

Dashron:
K, this is looking good, I just need to get mapping in place (simple) and keep slamming it with manual tests and throw together a bunch of unit tests. If you pull down and test with the polymorphism schema it should look right for the first case, which is fairly complex but structure wise pretty common. 

`nested-one-of-object-with-nested-one-of` on the other hand is broken, and needs a bit more investigation.

One thing I noticed while fixing this, is that the issue with multiselects not showing correctly on initial load seems to be something to do with the `<div className="children">` element not properly wrapping the children. I haven't looked further than that yet.

My planned tests are
- oneOf list of schemas with top level discriminator
- oneOf list of schemas with mappings and top level discriminator
- oneOf list of schemas with embedded discriminator
- oneOf list of schemas with mappings and embedded discriminator
- oneOfs with no discriminators, but one of the oneOfs has children oneOfs using a discriminator
- *maybe* examples of each of the above using anyOf in place of oneOf for some of the above

Dashron:
This is probably good for another review pass. The functionality is there, and multischemafield has tests in the api explorer.

Allowing conditional chaining highlighted a slew of hidden eslint objections because it considered that a parsing error, so I need to clean those up tomorrow. The main ones are my use of `for(... of ...)`, class functions that don't use `this` and arrow functions inside of classes (see notifySelectionValue and onOptionChange, which was a style used by the original MultiSchemaField).

I would also like to get some tests in on the changes to oas-form, but those are fairly minor.

packages/api-explorer/__tests__/form-components/MultiSchemaField.test.jsx
```diff
@@ -0,0 +1,197 @@
...
+const props = {
+  oas,
+  onChange: () => {},
+  onJsonChange: () => {},
+  onModeChange: () => {},
+  onSubmit: () => {},
+  resetForm: () => {},
+};
+
+function chooseOption(selectElement, optionNumber) {
```
Dashron, line 19: I love that I can do this so easily.

packages/api-explorer/src/form-components/MultiSchemaField.jsx
```diff
@@ -0,0 +1,299 @@
...
+  findEnumOptionsIndex(enumOptions, value) {
+    return enumOptions.findIndex((item) => item.value === value);
+  }
+
+  /**
+   * Notifies the onChange prop that the selected index has changed.
+   *
+   * Note that this is the index and not the value
+   */
+  notifySelectionValue = (selectedIndex, selectedValue) => {
```
Dashron, line 181: eslint doesn't like this

packages/api-explorer/src/form-components/MultiSchemaField.jsx
```diff
@@ -0,0 +1,299 @@
...
+    // Note: defaults might not be applicable, this line largely from the original multischema code!
+    onChange(getDefaultFormState(options[selectedIndex], formData, rootSchema));
+  }
+
+  /**
+   * Handles the dropdown selection of your schema.
+   *
+   * The selectedValue is the selection value not the index.
+   */
+  onOptionChange = (selectedValue) => {
```
Dashron, line 199: eslint doesn't like this

packages/api-explorer/src/form-components/MultiSchemaField.jsx
```diff
@@ -0,0 +1,299 @@
...
+
+/**
+ * Find the property that the discriminator is referencing.
+ */
+function findDiscriminatorField(schema, options, property) {
+  if (schema.properties?.[property]) {
+    return schema.properties[property];
+  }
+
+  for (const option of options) {
```
Dashron, line 19: eslint doesn't like me using iterators like this

packages/api-explorer/src/form-components/SchemaField.jsx
```diff
@@ -114,10 +115,16 @@ function SchemaField(props) {
...
   }
 
+  // If there's no name on this field, then it's a lone schema with no label or children and as such we shouldn't try
+  // to render it with the custom template.
   if ('name' in props) {
-    // If there's no name on this field, then it's a lone schema with no label or children and as such we shouldn't try
-    // to render it with the custom template.
-    props.registry.FieldTemplate = CustomTemplate;
+    // This allows an element to be created and bypass the template, for example when we hide the discriminator in the object ui
+    if (uiSchema['ui:customTemplate'] === false) {
```
Dashron, line 122: todo: test this

packages/oas-form/src/components/fields/StringField.js
```diff
@@ -22,12 +22,17 @@ function StringField(props) {
...
   const { widgets, formContext } = registry;
-  const enumOptions = isSelect(schema) && optionsList(schema);
+  let enumOptions = isSelect(schema) && optionsList(schema);
   let defaultWidget = enumOptions ? 'select' : 'text';
   if (format && hasWidget(schema, format, widgets)) {
     defaultWidget = format;
   }
   const { widget = defaultWidget, ...options } = getUiOptions(uiSchema);
+  // Allow the parent to override the options to provide custom keys via uiSchema:{ui:options: options}
+  if (options.enumOptions) {
```
Dashron, line 32: todo: test this

packages/oas-form/src/components/widgets/SelectWidget.js
```diff
@@ -92,7 +92,7 @@ function SelectWidget(props) {
       required={required}
       value={typeof value === 'undefined' ? emptyValue : value}
     >
-      {!multiple && schema.default === undefined && <option value="">{placeholder}</option>}
+      {!multiple && schema.default === undefined && !hideEmpty && <option value="">{placeholder}</option>}
```
Dashron, line 95: todo: test this

packages/api-explorer/src/form-components/SchemaField.jsx
Dashron: this is covered by the updated snapshot

Dashron:
k, tests done. Have at it!

Learned a lot with this one, and it should all help the redesign go much quicker.

packages/api-explorer/src/form-components/MultiSchemaField.jsx
```diff
@@ -0,0 +1,313 @@
...
+function findDiscriminatorField(schema, options, property) {
+  if (schema.properties?.[property]) {
+    return schema.properties[property];
+  }
+
+  // I need to return early, so I can't .forEach, wish I could use a for-of loop!
+  // eslint-disable-next-line no-plusplus
+  for (let i = 0; i < options.length; i++) {
+    const option = options[i];
+    if (option.properties?.[property]) {
```
erunion, line 19: We'll need to load https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining in order for us to safely use this. You should be able to add it as a devDep in the root directory and then add it into the `plugins` array in `.babelrc`.

packages/api-explorer/src/form-components/MultiSchemaField.jsx
```diff
@@ -0,0 +1,319 @@
...
+ * See getSchemaOptions for more thorough details. This basicaly does that, but uses the aliases in the mapping as
+ * values instead of the schema keys
+ */
+function getMappingOptions(mapping, discriminatorSchemas, componentSchemas) {
+  const validSchemas = {};
+
+  // Get a list of valid schemas listed in the mapping and format it to look like its own component schema object, but
+  // with the mapping keys instead of real keys so we can reuse getSchemaOptions.
+  Object.keys(mapping).forEach(key => {
+    const match = mapping[key].match(/#\/components\/schemas\/([^/]+)/);
```
erunion, line 90: Do we need to worry about `#/components/parameters`, `#/components/requestBodies`, or any other of the [other supported ones](https://swagger.io/docs/specification/components/)?

erunion (changes requested):
"Wow" -Owen Wilson

packages/api-explorer/src/form-components/MultiSchemaField.jsx
Dashron: Discriminator can only map to schema objects, so we shouldn't have to worry about anything else.

packages/api-explorer/src/form-components/MultiSchemaField.jsx
Dashron: Made the change locally, how would I ensure it's working? Optional chaining was working for me without this plugin

erunion: You should be able to run `npm run build` and then look at the source to make sure that `?.` isn't present.

erunion: Talked offline and the optional chaining plugin is now just part of `@babel/preset-env`.

Dashron: 👍  change pushed up

erunion (approved):

PR #1164, opened 2021-02-01
fix(ui components): remove a stylesheet import that's now redundant

kanadgupta:
| [☁️ &nbsp; CI App][demo] |
| --- |

## 🧰 What's being changed?

Now that we've migrated our full component library to the ReadMe repo, we need to remove a stylesheet import from the API Explorer repo since it's being applied twice. See https://github.com/readmeio/readme/pull/3552 for more info!

I verified that this doesn't break any of the styling in the demo app because of [`bundle-hub2.css`](https://github.com/readmeio/api-explorer/blob/0b3fbebdaefd2306e6fe393a1c05dc781f9ffa7e/example/bundle-hub2.css).

This also does a bit of recipes-specific housekeeping:
- [x] Updates the copy in the demo app so it's now "recipes" d90dc65
- [x] Fixes some of the logic so the "Enable recipes" toggle is now working again in the demo app. **Feel free to revert this if the toggle was intentionally disabled!** 0b3fbeb

## 🧬 Testing

Verify that the "Enable recipes" toggle in the demo app is working, and that none of the styling appears to be broken.

### DIFF
--- a/packages/api-explorer/src/Doc.jsx
+++ b/packages/api-explorer/src/Doc.jsx
@@ -10,7 +10,6 @@ const Oas = require('oas/tooling');
 const { getPath, matchesMimeType } = require('oas/tooling/utils');
 
 const { TutorialTile } = require('@readme/ui/.bundles/es/ui/compositions');
-require('@readme/ui/.bundles/umd/main.css');
 
 const isAuthReady = require('./lib/is-auth-ready');
 
--- a/example/src/Demo.jsx
+++ b/example/src/Demo.jsx
@@ -156,7 +156,7 @@ class Demo extends React.Component {
     if (enableTutorials) {
       docs.forEach((doc, i) => {
         // Only add our mock tutorials if we don't have them present.
-        if (!('tutorials' in doc)) {
+        if ('tutorials' in doc) {
           docs[i].tutorials = [
             {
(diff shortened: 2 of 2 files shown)


packages/api-explorer/src/Doc.jsx
```diff
@@ -10,7 +10,6 @@ const Oas = require('oas/tooling');
 const { getPath, matchesMimeType } = require('oas/tooling/utils');
 
 const { TutorialTile } = require('@readme/ui/.bundles/es/ui/compositions');
-require('@readme/ui/.bundles/umd/main.css');
```
rafegoldberg, line 13: This is definitely in keeping with the other CSS optimizations we've been making! But will be difficult to test until we run an integration, and I know we've had [some](https://github.com/readmeio/readme/blob/0fb6ae74028f78bfec5fec96208cb11d5b2a4879/packages/hub/public/css/bundle-hub2.styl#L74-L86) issue with cross-repo style application before.

rafegoldberg (approved):

erunion:
Looks good on the CI app: https://readme-api-e-fix-remove-qbdqgq.herokuapp.com/

erunion (approved):

PR #1165, opened 2021-02-02
Format useragent in simple terms

ilias-t:
| [☁️ &nbsp; CI App][demo] |
| --- |

## 🧰 What's being changed?

Certain user agents are simplified in terms of their naming & have specific icons to denote them:
image


package.json
```diff
@@ -50,6 +50,7 @@
     "@commitlint/cli": "^11.0.0",
     "@commitlint/config-conventional": "^11.0.0",
     "@readme/eslint-config": "^4.0.0",
+    "@svgr/webpack": "^5.5.0",
```
rafegoldberg, line 53: Any specific reason you like this @svgr/webpack loader over other ones?

packages/api-logs/main.css
```diff
@@ -104,6 +104,15 @@
 .hub-reference .logs .logs-list table th.time {
   width: 15%;
 }
+.hub-reference .logs .logs-list table td.useragent {
+  display: flex;
```
rafegoldberg, line 108: In general, we probably shouldn't be overriding the built-in  display of natural HTML elements, like forcing this `<td>` to display as a `flex` box. (Though I understand the reasoning behind your choice, for sure!) Hardly a showstopping issue, but if I had my druthers, I'd probably to wrap this in an extra `<span>` that I could display as an `inline-flex` box.

packages/api-logs/__tests__/utils.test.js
```diff
@@ -0,0 +1,55 @@
+import { getHeaderValue, getFormattedUserAgent } from '../utils';
+
+describe('API Logs Utils', () => {
```
rafegoldberg, line 3: Nice 👍

packages/api-logs/__tests__/utils.test.js
```diff
@@ -0,0 +1,55 @@
...
+    it('handles readme api explorer', () => {
+      expect(getFormattedUserAgent('ReadMe-API-Explorer')).toBe('ReadMe API Explorer');
+    });
+
+    it('handles node-fetch', () => {
+      expect(getFormattedUserAgent('node-fetch/x.x.x')).toBe('node');
+    });
+
+    it('handles python-request', () => {
+      expect(getFormattedUserAgent('python-request/x.x.x')).toBe('python');
```
rafegoldberg, line 48: I feel like we should capitalize these "pretty-printed" user agent strings if we're going through the trouble to iconize 'em? am I crazy?

rafegoldberg (approved):
The code looks good to me, barring some minor comments/nitpicking. (Though I should note I haven't tested this locally, or seen a ReadMe-side integration yet!)

packages/api-logs/__tests__/utils.test.js
rafegoldberg: (@yishinae might have opinions, if she's around!?)

yishinae: I'm indeed around! Although I'm not sure if I fully understand what we're talking about capitalizing!

rafegoldberg: The text next to the green icon in [this screengrab](#issue-565580862). But...........now that I look again it seems this already _is_ capitalized via CSS or something, so I guess this is really a non-issue. Marking this as resolved, but now that I know you're around expect many grammar-related tags!

yishinae: I'll be here!

package.json
ilias-t: The integration was pretty simple & had a good number of stars/downloads. It'd be pretty easy to switch it out if there's ever an issue with tit.

packages/api-logs/main.css
ilias-t: sure, works for me 👍

erunion:
Bundle size also jumped up a bit with this. Feel free to bump the threshold in the root `package.json`.

```
FAIL packages/api-logs/dist/index.js: 13.65KB > 10KB (gzip)
```

erunion (changes requested):
To show off each of these different language states, can you update https://github.com/readmeio/api-explorer/blob/next/webpack.config.js#L16-L22 to swap out different user agents for each fake log we return?

The bundle size also jumped up a bit with this, so feel free to bump the threshold in the root `package.json`.

```
FAIL packages/api-logs/dist/index.js: 13.65KB > 10KB (gzip)
```

mjcuva (approved):

PR #1166, opened 2021-02-02
fix: prevent code snippets from double encoding querystrings

erunion:
| [☁️ &nbsp; CI App][demo] |
| --- |

## 🧰 What's being changed?

Fixes a bug where if a query string had a `style` attribute to specify an encoding, it would end up getting doubly encoded between `@readme/oas-to-har` and `@readme/httpsnippet`.

Fixes RM-216

## 🧬 Testing

Check out the attached unit test. `2019-06-13T19:08:25.455Z` and `2015-09-15T14:00:12-04:00` was entered by the user and according to the API definition should be encoded with `form` styling. This data gets saved as `2019-06-13T19%3A08%3A25.455Z` into the HAR and fed into `@readme/oas-to-snippet` and `@readme/httpsnippet`.

`@readme/httpsnippet` should not encode this data again into `2019-06-13T19%253A08%253A25.455Z`.

[demo]: https://readme-api-e-fix-querys-rdllnu.herokuapp.com/

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5486,9 +5486,9 @@
       "integrity": "sha512-sC94wCjHfHYt87j+pKnG6FaKMX0N6BLBINPokR3XXGv43lhT32qbFbWDzmBuiJOC4PRHg0M4kMsBToVH47V6mA=="
     },
     "@readme/httpsnippet": {
-      "version": "2.3.1",
-      "resolved": "https://registry.npmjs.org/@readme/httpsnippet/-/httpsnippet-2.3.1.tgz",
-      "integrity": "sha512-gnL3cXfY+sG03SL/Ua5bEeiDxk+nTMMJu7+aKNLGZguL9R67vXvqL+A6mYaGwwUo6b7g5dfUzv0Zs44tfFQD1w==",
+      "version": "2.4.0",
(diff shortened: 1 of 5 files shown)


Dashron:
I noticed two things on https://readme-api-e-fix-querys-rdllnu.herokuapp.com/?selected=swagger-files%2Fparameter-styles-query.json

1. I think it's still double encoding? see ss in the "Demonstrate all non-exploded form query styles" section:
Screen Shot 2021-02-02 at 1 19 24 PM

2. I also noticed a bug that happens in production, so just calling it out. If you add a value to an object then remove it, it's replaced with `undefined`. Added an issue here: https://linear.app/readme-io/issue/RM-229/api-explorer-examples-are-incorrect-for-styled-query-objects.

erunion:
Yeah looks like you're right. Looking into it.

Screen Shot 2021-02-02 at 10 49 12 AM

erunion:
This is bugged out in the browser because Webpack 4 is injecting a `querystring` polyfill that doesn't have the same API as Node's `querystring`.

https://github.com/readmeio/httpsnippet/pull/35

erunion:
Fixed on the PR app:

kanadgupta (approved):
Confirmed in the PR app that query strings aren't double encoded:
Screen Shot 2021-02-03 at 11 26 34 AM

PR #1169, opened 2021-02-05
ci: experimenting with lighthouse

erunion:


lighthouserc.js
```diff
@@ -0,0 +1,26 @@
...
+      //   disableDeviceEmulation: true,
+      //   disableCpuThrottling: true,
+      //   disableNetworkThrottling: true,
+      //   throttlingMethod: 'provided',
+      // },
+    },
+    upload: {
+      target: 'lhci',
+      serverBaseUrl: process.env.LHCI_URL,
+      token: '8a40d86e-1b1d-4e69-b852-4eec88c37ffd', // process.env.LHCI_TOKEN,
```
domharrington, line 19: Is this okay to be public?

erunion: All anyone can do with it is submit bad lighthouse runs. Admin token is the one that really matters.

PR #1179, opened 2021-02-09
perf: oas-form bloat reduction and bundle overhauls

erunion:
[☁️ &nbsp; CI App][demo] | 🎫 RM-354
:---:|:---:

## 🧰 What's being changed?

* [x] Eliminate the need for `core-js` and instead use our core Babel config when building the dist for `@readme/oas-form`.
  * Since we're already loading in `@babel/preset-env` for handling polyfills, we don't need to also load in `core-js` for the same (and also we were only loading that in for a polyfill of `Array.includes` and `Array.fill`).
* [x] Stop building the `lib/` directory for `@readme/oas-form`. It's holdover from RJSF and we use it, need it, nor do we need to keep the Babel infrastructure in this package to maintain it.
* [x] Remove existing (unused) validation support.
  * Removes all support for exposing errors from AJV validation including the `rawErrors`, and `errorSchema` props as well as the `onError` event handler.
* [x] Removing support for JSON Schema `dependencies` from `oas-form`.
  * OAS 3.0 doesn't support `dependencies` at all. OAS 3.1 supports a flavor of this through `dependentSchemas`, but we aren't there yet to be able to adequately support this so I'm pulling it as it adds a chunk of unused bloat.
* [x] Removing the `omitExtraData` and `liveOmit` props.
  * They're there to prevent data that's not in the form from getting into `formData` when `onSubmit` is called but as we neither set these to `true` or have any cases where form data would be present without having gone through the form it doesn't make sense to keep it.
* [x] Removing the deprecated `autocomplete` prop.
  * There's an `autocomplete` and `autoComplete` prop. The former is deprecated for the latter. We use neither, but being able to extend that down to the `Form` component is nice and we might use it in the future so I'm removing the deprecated version.
* [x] Removing the `withTheme` export.
  * It's present for people to add additional theme-driven props to the core `Form` component, but since we just use this stuff directly we don't need it around. Dropping it also allows us to remove the dependency on `react-is` for detecting `React.forwardRef` components.
* [x] Removing the `ui:order` UI schema property for customizing the order of object properties.
  * We've never used this, have no way of using it with the way we generate JSON Schema, and have no plans of using it. 
* [x] Remove Lodash
  * With the removal of the above code we can now stop using Lodash and its `union`, `pick`, `get`, and `isEmpty` methods. 
* [x] Remove AJV

## 🐳  Why remove AJV and validation support?

The amount of overhead for what AJV does, validate JSON Schema, is incredible compared to what we actually need it to do: validate form inputs.

Since we don't need to use AJV to validate that arrays are arrays, or objects are objects, because those schemas are already handled by way of the form construction, what's left for it to validate is schema patterns. Well thankfully input patterns are natively supported on every browser that we target: https://caniuse.com/input-pattern. Now while we can't natively render multiple errors on inputs at the same time, we can use native APIs to compile those errors together and render them somewhere ([as demonstrated here](https://www.tjvantoll.com/2012/08/05/html5-form-validation-showing-all-error-messages/)).

That said, since we aren't even doing validation right now, and don't plan to do it with AJV in the future, it doesn't make sense to ship all this unused bloat right now and for the indefinite future.

| Branch | `@readme/api-explorer` | `@readme/oas-form` |
| :--- | :--- | :--- |
| Before (`next`) | 2.65 MiB | 297 KiB |
| After  | 2.46MiB | 136 KiB |

But wait, how come removing AJV didn't clear off hella space on the bundle? Well the short of it is that there's still a lot of packages that we're loading into the bundle that we shouldn't be; namely `@readme/syntax-highlighter`:

Screen Shot 2021-02-17 at 4 35 20 PM

## 🧙  Future

There are still a lot of improvements that we can make to further reduce the size of our API Explorer bundles:

* Treat `@readme/syntax-highlighter` as an external. Unfortunately this isn't that easy because `@readme/ui` and `@readme/oas-to-snippet` load it in so we'd need to make some improvements there.
* Since `@readme/ui` is deprecated we should pull in the components that we use there into this library. This'll allow to exclude all components, compositions, and views that we don't use.
* Possibly fork [`@apidevtools/json-schema-ref-parser`](https://www.npmjs.com/package/@apidevtools/json-schema-ref-parser)? It has support for parsing YAML files, but since our `oas` library, which loads it for dereferencing, only handles JSON, it doesn't make much sense to ship full copies of `js-yaml` and `esprima` in our bundles.

## 🧬 Testing

* [ ] Does the explorer throw any erorrs?
* [ ] Does [polymorphism](https://readme-api-e-refactor-o-hzudqu.herokuapp.com/?selected=swagger-files%2Fpolymorphism.json) still work?
* [ ] Do discriminators still work?
* [ ] Does form data updating still work?
* [ ] Does Try It still work?

[demo]: https://readme-api-e-refactor-o-hzudqu.herokuapp.com/


packages/api-explorer/webpack.config.js
```diff
@@ -18,6 +20,11 @@ module.exports = {
       amd: 'react-dom',
       umd: 'react-dom',
     },
+
+    // `@readme/ui` loads this in for the search components but unfortunately we aren't able to exclude this from being
+    // compiled into our dist, despite not using that component. Tf we treat it as a Webpack external it will thankfully
+    // be ignored.
+    'react-instantsearch-dom': 'react-instantsearch-dom',
```
erunion, line 27: This feels a little sketchy, but it seems to work?

rafegoldberg: lol this is nothing as far as ridiculous webpack `externals` go. check out [the maniacal shit](https://github.com/readmeio/readme/blob/54cd7fe31736b82400f879b76b93583a33a7d5b1/packages/react/webpack.config.js#L14-L43) I'm doing in the React subpackage...

rafegoldberg: By which I mean I believe this is generally the correct way to externalize. Can we not do something like this for our syntax highlighter? Or is that verboten cause deps of deps rely on it? Oy vey...

rafegoldberg (commented):
This work looks great, and from my brief tests...

- [x] **Does the explorer throw any erorrs?**
  Randomly poked through our list of OAS schemas in the CI app. Most rendered great; where I did see an endpoint or two erroring out I checked em against "production" (preview.readme.io, that is) and they were also erroring there.
- [x] **Does polymorphism still work?**
  I checked this out [the polymorphism example](https://readme-api-e-refactor-o-hzudqu.herokuapp.com/?selected=swagger-files%2Fpolymorphism.json) in PR app and played around with updating various data in the form and editor. E'erything seemed in good working order, though I'm an OAS noob, so...
- [x] **Does form data updating still work?**
  Yep, snippet generation looks good. JSON editor reset is working nicely.

Gonna leave some of these to those more qualified! 😶

- [ ] ~Does Try It still work?~
- [ ] ~Do discriminators still work?~

rafegoldberg (approved):

packages/api-explorer/webpack.config.js
erunion: @rafegoldberg Yeah both the explorer and `@readme/ui` require `@readme/oas-to-snippet` which uses the syntax highlighter. I'm going to play around with another PR to try to remove it from the dist, but it seemed a bit much on top of everything I'd already done here.

erunion: @rafegoldberg Since `@readme/ui` is deprecated, thoughts on us forking the tutorial components we're using from it into this repo?

erunion:
> Randomly poked through our list of OAS schemas in the CI app. Most rendered great; where I did see an endpoint or two erroring out I checked em against "production" (preview.readme.io, that is) and they were also erroring there.

@rafegoldberg which ones did you find errors on?

rafegoldberg:
@erunion there were two I remember, but [this is the only one](https://preview.readme.io/?selected=https%3A%2F%2Fapi.apis.guru%2Fv2%2Fspecs%2Famazonaws.com%2Fworkspaces%2F2015-04-08%2Fopenapi.json) I still have the URL of on my clipboard.

packages/api-explorer/webpack.config.js
rafegoldberg: Honestly, the thought of forking the tutorial component makes me shudder. The [component architecture](https://github.com/readmeio/readme/tree/next/packages/react/src/routes/Tutorials/components) is quite complex, with a not-insignificant [number of](https://github.com/readmeio/readme/blob/6cdf46723176694ffccc3701826811bfaa1c6276/packages/react/src/core/hooks/useReadmeApi.js) [cross-dependencies](https://github.com/readmeio/readme/blob/0c2509b20e3e18de02246e374c438aee593d9208/packages/react/src/ui/Modal/index.js#L0-L1)... Plus, post-migration (🎫 RM-219), this component _already_ lives in two places!

Considering the [big rewrite](https://linear.app/readme-io/project/20b9da3b992e) we're about to undertake (🎫 RM-246, specifically) I think my line here is "why kill ourselves doing now what we'll just have to redo/refigure a month or two down the line?" Plus I think the colocation of the Explorer and the React app will redound to our benefit in these exact kind of "component sharing" scenarios! In short, I'd hold off on this till we're further in to the refactor.

erunion: Ended up ripping UI components out of the Explorer completely and are passing them in as props now. #819

packages/oas-form/src/components/fields/MultiSchemaField.js
```diff
@@ -1,57 +1,17 @@
...
+import { getUiOptions, getWidget, guessType, retrieveSchema, getDefaultFormState } from '../../utils';
 
 class AnyOfField extends Component {
   constructor(props) {
     super(props);
 
-    const { formData, options } = this.props;
-
     this.state = {
-      selectedOption: this.getMatchingOption(formData, options),
```
Dashron, line 21: Do we ever provide form data to an initial page load? If so, losing this will stop the polymorphism dropdown from having an initially selected value

packages/oas-form/src/utils.js
```diff
@@ -246,9 +242,6 @@ function computeDefaults(_schema, parentDefaults, rootSchema, rawFormData = {},
     }
 
     return computeDefaults(refSchema, defaults, rootSchema, formData, includeUndefinedValues);
```
Dashron, line 244: Discriminators uses this, but I never really bothered to check what it was for? If the discriminator tests run I probably feel fine with this change.

packages/oas-form/src/utils.js
```diff
@@ -691,124 +648,6 @@ export function retrieveSchema(schema, rootSchema = {}, formData = {}) {
   return resolvedSchema;
 }
 
-function resolveDependencies(schema, rootSchema, formData) {
```
Dashron, line 694: What are dependencies and why don't we need this?

packages/oas-form/src/utils.js
erunion: This is what compiles the original default `formData` on load.

packages/oas-form/src/components/fields/MultiSchemaField.js
erunion: Thankfully we don't. Only data we provide is through JSON Schema `default` props, and those get automatically compiled into the initial page load through `computeDefaults`.

packages/oas-form/src/utils.js
erunion: That's all from this https://react-jsonschema-form.readthedocs.io/en/latest/usage/dependencies/. It's not part of the JSON Schema spec anymore.

PR #1180, opened 2021-02-10
fix: missing operation schema exceptions

erunion:
| [☁️ &nbsp; CI App][demo] |
| --- |

## 🧰 What's being changed?

* [x] Updates our `oas` dependency to fix a bug where if a supplied path+method combination didn't exist in an OAS it was being searched for in though we'd still return an instance of the `Operation` class, we'd return it with an `undefined` `schema` property. This error would then cascase and throw failures on any accessor of that class. Check out https://github.com/readmeio/oas/pull/371 for tests.

## 🧬 Testing

Check out the attached `manual-api-missingInOas.json` example definition.

Against `next`, this is what it currently looks like:

Screen Shot 2021-02-09 at 6 56 39 PM

And with this fix:

Screen Shot 2021-02-09 at 6 58 14 PM

### DIFF
--- a/example/swagger-files/manual-api-missingInOas.json
+++ b/example/swagger-files/manual-api-missingInOas.json
@@ -1,0 +1,54 @@
+{
+  "readmeManual": true,
+  "docs": [
+    {
+      "_id": "7333016a-a191-4722-ba46-be45a19ddf6c",
+      "swagger": {
+        "path": "/",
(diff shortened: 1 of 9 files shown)


rafegoldberg (approved):
Not as looped in on this, but the code looks good to me!

billhimmelsbach (approved):
Looking good!

Screen Shot 2021-02-09 at 7 03 13 PM

PR #1181, opened 2021-02-11
feat: start exposing `pattern` attributes in schemas

erunion:
| [☁️ &nbsp; CI App][demo] |
| --- |

## 🧰 What's being changed?

* [x] If there is a `pattern` attribute in a schema, we'll now add that pattern into the `input` element for that parameter.

## 🧬 Testing

See the `types.json` example and the accompanying unit test.


Dashron (approved):
simple enough

PR #1189, opened 2021-02-18
feat: populate params with JWT defaults

darrenyong:
| [☁️ &nbsp; CI App][demo] |
| --- |

## 🧰 What's being changed?
With the change to `Operation.getParametersAsJsonSchema()` in the [oas library](https://github.com/readmeio/oas/pull/373), it accepts a new object which adds support to allow users to define parameter defaults which can be passed in through the JWT User Data object.

Since the data will come in in the following format, we can key directly into `.variables.user.parameters` and pass that through to the Params component, which will then get passed to `Operation.getParametersAsJsonSchema()`
```
{
  name: "Marc Cuva",
  email: "marc@readme.io",
  ...,
  parameters: {
    id: 5,
    name: Owlbert,
    version: 1.0
  }
}
```

## 🧬 Testing

In Demo.jsx, there's a `variables.user` object. 
1. Add a `parameters` key object within `variables.user`. 
2. In `parameters`, add keys that correspond with parameter names. 

image
For example, if we wanted to match all the parameters in the endpoint above, we would add the following:
```
parameters: {
  id: 24,
  category: {
    id: 49,
    name: 'Owlbert'
  },
  ...,
}
```
3. Adding parameters that don't correspond to parameters shouldn't add defaults. So if you add the following, it shouldn't show up anywhere.
```
parameters: {
  fakeParameter: 'testing'
}
```


packages/api-explorer/src/Params.jsx
```diff
@@ -195,6 +195,7 @@ Params.propTypes = {
   onModeChange: PropTypes.func.isRequired,
   onSubmit: PropTypes.func.isRequired,
   operation: PropTypes.instanceOf(Operation).isRequired,
+  parameters: PropTypes.object,
```
erunion, line 198: Can you add a default empty object to the `defaultProps` for this?

packages/api-explorer/src/Doc.jsx
```diff
@@ -481,6 +481,7 @@ class Doc extends React.Component {
         onModeChange={this.onModeChange}
         onSubmit={this.onSubmit}
         operation={this.getOperation()}
+        parameters={this.props.user.parameters}
```
erunion, line 484: `parameters` likely won't always be there so this'll come in as `undefined`. Could maybe write this as:

```suggestion
        parameters={this.props.user?.parameters || {}}
```

package-lock.json
```diff
@@ -152,58 +152,11 @@
         "z-schema": "^4.2.3"
       }
     },
-    "@babel/cli": {
```
erunion, line 155: Can you re-run `npm run bootstrap` from the root? It might clean up some of these lockfile changes.

erunion:
@mjcuva What're your thoughts on adding these JWT defaults for parameters into a `parameters` key in the JWT? I talked with @domharrington yesterday and he thought it might be better if they were in the `keys` array so individual keys could specify their own defaults, and while I like that after sleeping on it I'm not sure how we'd expose or update those defaults when you swap API keys.

domharrington:
I think there's 2 points here:
- should they be in the keys array
- should they be under a new key, or on the top level

I chatted with @mjcuva about this already! I personally think they should be in the keys array as that matches how other things work... a key is almost like a user.

On the 2nd point, I dont have many opinions. It may be slightly weird if we were populating our required JWT keys in to fields marked with id/apiKey/email?

domharrington:
One other question about implementation on this - how does it work for nested keys? Can I provide a nested object of params? Does that populate a nested object form?

domharrington (approved):
The code here looks relatively minimal, i'm guessing the bulk of the actual logic occurs in OAS itself? 

Do we need any integration tests in this repo?

Dashron (approved):

jordanmnunez:
mixpanel dev docs are v excited for this

darrenyong:
> I think there's 2 points here:
> 
> * should they be in the keys array
> * should they be under a new key, or on the top level
> 
> I chatted with @mjcuva about this already! I personally think they should be in the keys array as that matches how other things work... a key is almost like a user.
> 
> On the 2nd point, I dont have many opinions. It may be slightly weird if we were populating our required JWT keys in to fields marked with id/apiKey/email?

I was thinking we can eventually allow people to add it into the `keys` object. This way, if a person has one project, they can add it into the root level, but if they have multiple projects, they can add it into the `keys` object!

> Great job! 💪 The code here looks relatively minimal, i'm guessing the bulk of the actual logic occurs in OAS itself?
> 
> Do we need any integration tests in this repo?



> One other question about implementation on this - how does it work for nested keys? Can I provide a nested object of params? Does that populate a nested object form?

Yep! The logic is in this PR [here](https://github.com/readmeio/oas/pull/373)! And it should work for infinitely nested objects.

erunion:
This is going to require a bit of reworking within the redesign code. @darrenyong you want to move this PR over to there?

darrenyong:
> This is going to require a bit of reworking within the redesign code. @darrenyong you want to move this PR over to there?

Yep! I'll recreate this on the redesign code!

erunion:
We're reworking this into the redesign so I'm closing this out.

PR #1192, opened 2021-02-18
test: new test to cycle and check all oas examples for errors

Dashron:
| [☁️ &nbsp; CI App][demo] |
| --- |

## 🧰 What's being changed?

This adds a test that scans a folder full of oas files and renders them in the code. If any of them contain the "An unexpected error" message, or throws a react warning, it will fail the test.

This is also useful locally to run on a much larger set of files.

In the future, we should continue to expand our examples so that we're testing a wider range of functionality.


erunion (approved):
```
@readme/api-explorer: PASS __tests__/test-all-oas.test.jsx
```

noice

Dashron:
Moved this file into the new explorer at https://github.com/readmeio/readme/pull/3761. I don't think we need to merge this into the old one.

PR #1193, opened 2021-02-18
refactor: decoupling our ui library from the api explorer codebase

erunion:
| [☁️ &nbsp; CI App][demo] |
| --- |

## 🧰 What's being changed?

Last month we deprecated `@readme/ui`, but unfortunately our API Explorer still ships with it. This work fully decouples that library from this codebase, which'll allow to keep the components that we use here up to date with while also allowing us to shave a nice chunk off the bundle size.

Now when we instantiate the `APIExplorer` component in ReadMe we'll need to pass it a `ui` prop containing the following:

```js
{
  // @readme/ui/.bundles/es/ui/components/Button
  Button, 

  // @readme/ui/.bundles/es/ui/components/Tabs
  Tabs,

  tutorials: {
    // @readme/ui/.bundles/es/ui/compositions/Tutorials/Modal/constants/stepDefaults.DEFAULT_TUTORIAL
    DEFAULT_TUTORIAL, 
    
    // @readme/ui/.bundles/es/ui/compositions/TutorialModal
    TutorialModal,
    
    // @readme/ui/.bundles/es/ui/compositions/TutorialTile
    TutorialTile, 

    // @readme/ui/.bundles/es/views/cmVariableContext
    TutorialVariableContext 
  }
}
```

However since we still have unit tests that need to do determinations based on these components, and https://preview.readme.io still has to work `@readme/ui` will remain as a dev dependency here -- it just won't be shipped with our bundle.

## 🧹 Bundle size changes

> The file sizes here are what Webpack outputs to the console when running `npm run build`. They do not take gzipping into account.

| Branch | `@readme/api-explorer` |
| :--- | :--- |
| Before (`next`) | 2.65 MiB |
| After  | 2.24MiB |

## 🧬 Testing

* [ ] Does the JSON editor still work?
* [ ] Do the recipe tiles and modals still work?

[demo]: https://explorer-pr-1193.herokuapp.com/

### DIFF
--- a/example/src/Demo.jsx
+++ b/example/src/Demo.jsx
@@ -1,7 +1,9 @@
 const React = require('react');
 const PropTypes = require('prop-types');
 
-const { Button, Flex, Tabs, Toggle, TutorialModal, TutorialTile } = require('@readme/ui/.bundles/es/ui/components');
+const { Button, Flex, Tabs, Toggle } = require('@readme/ui/.bundles/es/ui/components');
+const { TutorialModal, TutorialTile } = require('@readme/ui/.bundles/es/ui/compositions');
+
 const { cmVariableContext: TutorialVariableContext } = require('@readme/ui/.bundles/es/views');
 const { DEFAULT_TUTORIAL } = require('@readme/ui/.bundles/es/ui/compositions/Tutorials/Modal/constants/stepDefaults');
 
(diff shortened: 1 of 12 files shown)


kanadgupta (approved):
Code looks good, and confirmed that all the components in question operate as expected in the PR app!

PR #1197, opened 2021-02-24
fix(oas-to-har): clean up formatting of raw json bodies

Dashron:
| [☁️ &nbsp; CI App][demo] |
| --- |

## 🧰 What's being changed?

When your request body has a type of `string` and format of `json`, the exact value was showing up as the user inputted string (with any included formatting (such as newlines)), passed through JSON.stringify .

e.g.

`--data='"{\n    \"foo\": \"bar\"\n}"'`

This changes that field to now appear as a JSON object passed thorugh JSON.stringify

e.g.
`--data='{"foo":"bar"}'`

## 🧬 Testing

Check the tests, or message me and I'll send you an example OAS file.

### DIFF
--- a/packages/oas-to-har/src/index.js
+++ b/packages/oas-to-har/src/index.js
@@ -303,33 +303,43 @@ module.exports = (
           } else {
             har.postData.mimeType = contentType;
 
-            // Find all `{ type: string, format: json }` properties in the schema because we need to manually JSON.parse
-            // them before submit, otherwise they'll be escaped instead of actual objects.
-            // We also only want values that the user has entered, so we drop any undefined cleanBody keys
-            const jsonTypes = Object.keys(requestBody.schema.properties).filter(
-              key => requestBody.schema.properties[key].format === 'json' && cleanBody[key] !== undefined
-            );
+            // Handle arbitraty JSON input via a string.
+            // In OAS you usually find this in an application/json content type.
+            //   with a schema type=string, format=json.
+            // In the UI this is represented by an arbitrary text input
+            // This ensures we remove any newlines or tabs and use a clean json block in the example
+            if (requestBody.schema.type === 'string') {
+              har.postData.text = JSON.stringify(JSON.parse(cleanBody));
+            } else {
+              // Handle formatted JSON objects that have properties that accept arbitrary JSON
+              // Find all `{ type: string, format: json }` properties in the schema because we need to manually JSON.parse
+              // them before submit, otherwise they'll be escaped instead of actual objects.
+              // We also only want values that the user has entered, so we drop any undefined cleanBody keys
+              const jsonTypes = Object.keys(requestBody.schema.properties).filter(
+                key => requestBody.schema.properties[key].format === 'json' && cleanBody[key] !== undefined
+              );
 
-            if (jsonTypes.length) {
-              try {
-                jsonTypes.forEach(prop => {
-                  try {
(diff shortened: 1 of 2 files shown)


Dashron:
Note: these fields still have a strange error. If you empty out a request body of type="string" and format="json" it's replaced immediately with the default value, making it difficult to work with. We should tackle that in the future.

packages/oas-to-har/src/index.js
```diff
@@ -303,33 +303,43 @@ module.exports = (
...
-            const jsonTypes = Object.keys(requestBody.schema.properties).filter(
-              key => requestBody.schema.properties[key].format === 'json' && cleanBody[key] !== undefined
-            );
+            // Handle arbitraty JSON input via a string.
+            // In OAS you usually find this in an application/json content type.
+            //   with a schema type=string, format=json.
+            // In the UI this is represented by an arbitrary text input
+            // This ensures we remove any newlines or tabs and use a clean json block in the example
+            if (requestBody.schema.type === 'string') {
+              har.postData.text = JSON.stringify(JSON.parse(cleanBody));
```
Dashron, line 312: The diff sucks on this. Really all I added was this line, and indented everything else that existed under the else.

The problem was that line 318 in the right hand file was failing because these schemas have no properties, and Object.keys would error on undefined.

packages/oas-to-har/src/index.js
```diff
@@ -303,33 +303,43 @@ module.exports = (
...
-            // We also only want values that the user has entered, so we drop any undefined cleanBody keys
-            const jsonTypes = Object.keys(requestBody.schema.properties).filter(
-              key => requestBody.schema.properties[key].format === 'json' && cleanBody[key] !== undefined
-            );
+            // Handle arbitraty JSON input via a string.
+            // In OAS you usually find this in an application/json content type.
+            //   with a schema type=string, format=json.
+            // In the UI this is represented by an arbitrary text input
+            // This ensures we remove any newlines or tabs and use a clean json block in the example
+            if (requestBody.schema.type === 'string') {
```
erunion, line 311: Do you want to add a check to make sure `type` exists in `requestBody.schema`?

Dashron: This essentially works that way right? The value will be undefined which isn't equal to `string`?

erunion: I've been burned so often on this shit so I'm always nervous about it, but you're right.

```
let tktk = {schema: {}}; tktk.schema.type === 'string'
> false
```

packages/oas-to-har/src/index.js
```diff
@@ -303,33 +303,43 @@ module.exports = (
           } else {
             har.postData.mimeType = contentType;
 
-            // Find all `{ type: string, format: json }` properties in the schema because we need to manually JSON.parse
-            // them before submit, otherwise they'll be escaped instead of actual objects.
-            // We also only want values that the user has entered, so we drop any undefined cleanBody keys
-            const jsonTypes = Object.keys(requestBody.schema.properties).filter(
-              key => requestBody.schema.properties[key].format === 'json' && cleanBody[key] !== undefined
-            );
+            // Handle arbitraty JSON input via a string.
```
erunion, line 306:
```suggestion
            // Handle arbitrary JSON input via a string.
```

erunion (approved):

packages/oas-to-har/src/index.js
Dashron: Arbitraty. Sigh.

PR #1240, opened 2021-03-24
feat: adding a new `validateExtension` method to `oas-extensions`

erunion:
## 🧰 What's being changed?

Ports over a new `validateExtension` method into `oas-extensions` that we've had in place in readme/readme, while also adding support for the new `x-readme` grouping.

I've also removed the `RESPONSE_SAMPLES` extension I added yesterday as we don't actually need it.


domharrington (approved):
LGTM!

PR #1250, opened 2021-04-16
fix: fatal error where basic auth scheme might be missing

erunion:
## 🧰 What's being changed?

This fixes a bug in `@readme/oas-to-har`'s `configure-security` library where if an OAS with a basic auth scheme does not exist within the provided auth data, and that auth data is **not** an empty object, we'd throw a fatal error when attempting to access `.user` on `undefined`

## 🧬 Testing

See tests


billhimmelsbach (approved):
Super appreciate the update to the tests to make them more relevant!

Now that we're failing more gracefully! Off to the useEffect! 🏃‍♂️

PR #1255, opened 2021-04-20
fix: removing undefined formData before processing it into postData

erunion:
## 🧰 What's being changed?

This fixes a bug with `application/x-www-url-formencoded` handling within `@readme/oas-to-har` where we weren't properly filtering out `undefined` values from `formData` before processing it into `postData` resulting in a case where if you added and removed data from the form we'd send `'undefined'` into the HAR (and thus into code snippets).

https://user-images.githubusercontent.com/33762/115469756-1617a680-a1ea-11eb-8f0c-abb48ef60e2c.mov

## 🧬 Testing

You can see this in action on https://preview.readme.io/?selected=swagger-files%2Fformdata.json

With this fix:

https://user-images.githubusercontent.com/33762/115469860-3cd5dd00-a1ea-11eb-8f7e-35a9f4c2bb9f.mov

### DIFF
--- a/packages/oas-to-har/__tests__/index.test.js
+++ b/packages/oas-to-har/__tests__/index.test.js
@@ -1404,6 +1404,36 @@ describe('requestBody', () => {
       ).toBe(JSON.stringify({ a: 'value' }));
     });
 
+    it('should not add undefined formData into postData', () => {
+      expect(
+        oasToHar(
+          oas,
+          {
+            path: '/',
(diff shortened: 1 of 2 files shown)


Dashron (approved):
code looks fine, will test when pr is up

erunion:
Confirmed on the pr app https://explorer-pr-1255.herokuapp.com/?selected=swagger-files%2Fformdata.json

PR #1257, opened 2021-04-21
feat: adding support for responses that don't have examples

erunion:
| [☁️ &nbsp; CI App][demo] |
| --- |

## 🧰 What's being changed?

* [x] Updates `oas` to pull in support for response examples that don't actually have an example (like `204 No Content`). https://github.com/readmeio/oas/pull/410
* [x] Updates our `ResponseExample` component to support these example-less responses.
  * In leu of rendering out example content for these responses that we obviously don't have we're going to instead render out the message that accompanies that HTTP status code. So if it's a 204 response the example will be "No Content". If the status code is unrecognized we'll render `HTTP Status Code: <status code>`

## 🧬 Testing

This is what operations with responses that don't have examples looks like currently:

https://user-images.githubusercontent.com/33762/115635636-0a4ae380-a2c1-11eb-8d32-c8e165269c0a.mov

And what it'll now look like:

https://user-images.githubusercontent.com/33762/115635657-12a31e80-a2c1-11eb-82a0-a0d65da31e17.mov

Provide as much information as you can on how to test what you've done.

[demo]: https://explorer-pr-1257.herokuapp.com/


example/swagger-files/examples.json
```diff
@@ -26,7 +26,7 @@
               }
             }
           },
-          "204": {
+          "206": {
```
ilias-t, line 29: 👍

packages/api-explorer/src/ResponseExample.jsx
```diff
@@ -215,6 +213,29 @@ class ResponseExample extends React.Component {
...
+                  let statusMessage;
+                  try {
+                    const code = getStatusCode(ex.status);
+                    statusMessage = code.message;
+                  } catch (err) {
+                    statusMessage = `HTTP Status Code: ${ex.status}`;
+                  }
+
+                  return (
+                    <div key={index}>
```
ilias-t, line 228: i know we're doing this below as well, but if there's anything more unique that we have than the index that'd be better

packages/api-explorer/src/ResponseExample.jsx
```diff
@@ -259,7 +280,7 @@ class ResponseExample extends React.Component {
           </span>
         )}
 
-        {(examples.length === 0 || (!hasExamples && result === null)) && (
+        {(examples.length === 0 || !result === null) && (
```
ilias-t, line 283: `!result === null` will never be true, I'm guessing this is supposed to be `result === null`. FWIW I also prefer `result == null` cause that's inclusive of both (and only) null & undefined. Lodash's [isNil](https://github.com/lodash/lodash/blob/4.17.15/lodash.js#L11974) is just that.

ilias-t (changes requested):

ilias-t:
I think there's a boolean logic error, which this only reason I didn't ✅

packages/api-explorer/src/ResponseExample.jsx
erunion: I'm going to remove this `result` check and the prop from this component completely since `examples.length` is always going to take precedence over this block now that we're handling empty responses.

packages/api-explorer/src/ResponseExample.jsx
erunion: Modified it to `response-{index}`

packages/api-explorer/src/ResponseExample.jsx
```diff
@@ -257,9 +278,7 @@ class ResponseExample extends React.Component {
               })}
             </div>
           </span>
-        )}
-
-        {(examples.length === 0 || (!hasExamples && result === null)) && (
```
erunion, line 262: Removing this `result` check from here since `examples.length` now takes precedence over this as we're now handling responses w/o examples.

ilias-t (approved):

PR #1258, opened 2021-04-23
fix: enable code copy buttons

kellyjosephprice:
| [☁️ &nbsp; CI App][demo] | Fix CX-281 |
| --- | --- |

## 🧰 What's being changed?

Enables the copy button on pinned code blocks.

## 🧬 Testing

You should be able to see and use the copy buttons on code blocks.

[demo]: https://explorer-pr-1258.herokuapp.com/?selected=swagger-files%2Fexcerpt-markdown.json


erunion:
Are there CSS fixes we need to put in place alongside this? They were originally removed in https://github.com/readmeio/api-explorer/pull/986 because they made code blocks look like:

96058734-e190ef80-0e40-11eb-969f-c45a9dab56a6

kellyjosephprice:
> Are there CSS fixes we need to put in place alongside this? They were originally removed in #986 because they made code blocks look like:
> 
> 96058734-e190ef80-0e40-11eb-969f-c45a9dab56a6

I didn't enable it for response schemas. So that _shouldn't_ be an issue.

kellyjosephprice:
From the CI App:
image

kellyjosephprice:
@domharrington and @erunion I think this is ready for review.

packages/api-explorer/src/block-types/Content.jsx
```diff
@@ -48,7 +48,7 @@ const Content = props => {
   if (useNewMarkdownEngine) {
     const content = markdown(body, {
       compatibilityMode: flags.rdmdCompatibilityMode,
-      copyButtons: false,
+      copyButtons: flags.copyButtons,
```
erunion, line 51: Is this a flag that we have on projects?

packages/api-explorer/src/block-types/Content.jsx
```diff
@@ -128,6 +128,7 @@ Loop.defaultProps = {
 
 Content.propTypes = {
   body: PropTypes.string,
+  copyButtons: PropTypes.bool,
```
ilias-t, line 131: nit: can we rename to `isCopyEnabled` or something?

ilias-t (approved):

packages/api-explorer/src/block-types/Content.jsx
kellyjosephprice: Yea, but to be consistent, I'll have to change it in markdown first, then api-explorer and readme.

ilias-t: ok up to you!

packages/api-explorer/src/block-types/Content.jsx
```diff
@@ -43,12 +43,12 @@ const Loop = ({ content, column, flags, splitReferenceDocs }) => {
...
 
 const Content = props => {
-  const { body, flags, isThreeColumn, splitReferenceDocs, useNewMarkdownEngine } = props;
+  const { body, copyButtons, flags, isThreeColumn, splitReferenceDocs, useNewMarkdownEngine } = props;
 
   if (useNewMarkdownEngine) {
     const content = markdown(body, {
       compatibilityMode: flags.rdmdCompatibilityMode,
-      copyButtons: false,
+      copyButtons,
```
erunion, line 51: Since this component is used in response schemas that are funneled through Markdown, this is going to add copy buttons to schema descriptions there. Can you set a `false` default to `copyButtons` so the only thing that'll get copy buttons are operation descriptions in the center column?

kellyjosephprice: https://github.com/readmeio/api-explorer/pull/1260

PR #1261, opened 2021-04-30
feat: adding support to oas-to-snippet for alternative language targets

erunion:
| [☁️ &nbsp; CI App][demo] |
| --- |

## 🧰 What's being changed?

This updates `@readme/oas-to-snippet` to add support alternative language targets! 

Previously if you wanted to generate a Node snippet, you'd do:

```js
generateCodeSnippet(oas, operation, formData, authData, 'node', oasUrl);
```

This would generate you a `node-fetch` snippet and that's great but [HTTPSnippet](https://npm.im/@readme/httpsnippet) offers all sorts of other Node targets like `axios` or `request` that you couldn't use &mdash; with this change you can now do the following:

```js
generateCodeSnippet(oas, operation, formData, authData, ['node', 'axios'], oasUrl);
```

With the `language` argument as an array we'll now do a lookup in our supported targets for `node` and if `axios` is there we'll generate an Axios snippet!

## 🧬 Testing

I've mostly rewritten `supportedLanguages` and the core language determination in this library along with a ton of new tests for this and all of the new targets we're now exposing, but with that there's a few points:

* Backwards compatibility remains intact so `node` will continue to map to `node-fetch`.
  * https://explorer-pr-1261.herokuapp.com/?selected=swagger-files%2Fpetstore.json
* I've also removed `node-simple` from the list of `supportedLanguages` in favor of treating it as a proper alternative target for `node`. To retain backwards compatibility with the rest of these changes, supplying `node-simple` as the language arg (and having it within your OAS in `x-samples-languages`) will continue to work.
  * https://explorer-pr-1261.herokuapp.com/?selected=swagger-files%2Fauth-types.json 
* I've dug through how we use `supportedLanguages` within ReadMe and thankfully the overhauls to the structure of this object won't affect anything there.

[demo]: https://explorer-pr-1261.herokuapp.com/

### DIFF
--- a/packages/oas-to-snippet/__tests__/index.test.js
+++ b/packages/oas-to-snippet/__tests__/index.test.js
@@ -3,6 +3,7 @@ const Oas = require('oas/tooling');
 const petstore = require('@readme/oas-examples/3.0/json/petstore.json');
 const path = require('path');
 const datauri = require('datauri');
+const harExamples = require('har-examples');
 
 const generateCodeSnippet = require('../src');
 const supportedLanguages = require('../src/supportedLanguages');
@@ -10,6 +11,7 @@ const supportedLanguages = require('../src/supportedLanguages');
 const { getLangName } = generateCodeSnippet;
 
 const oas = new Oas();
+const petstoreOas = new Oas(petstore);
 
 const oasUrl = 'https://example.com/openapi.json';
 const operation = {
@@ -27,28 +29,7 @@ const operation = {
 const formData = { path: { id: 123 } };
 
 test('should be able to accept a har override', () => {
-  const harOverride = {
-    log: {
-      entries: [
-        {
-          request: {
-            method: 'GET',
-            url: 'https://dash.readme.io/api/v1/categories/',
-            httpVersion: 'HTTPS/1.1',
-            headers: [
-              {
-                name: 'authorization',
-                value: 'Basic xxx',
-              },
-            ],
-            queryString: [],
-          },
-        },
-      ],
-    },
-  };
-
-  const codeSnippet = generateCodeSnippet(oas, operation, {}, {}, 'node', oasUrl, harOverride);
+  const codeSnippet = generateCodeSnippet(null, null, null, null, 'node', null, harExamples.full);
   expect(codeSnippet).toMatchSnapshot();
 });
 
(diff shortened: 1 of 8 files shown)


packages/oas-to-snippet/__tests__/index.test.js
```diff
@@ -27,28 +29,7 @@ const operation = {
...
-                name: 'authorization',
-                value: 'Basic xxx',
-              },
-            ],
-            queryString: [],
-          },
-        },
-      ],
-    },
-  };
```
erunion, line 49: Removed this hardcoded HAR in favor of using one that I copied from HTTPSnippet into https://github.com/readmeio/har-examples (we use this repo for testing `api` snippets).

packages/oas-to-snippet/__tests__/index.test.js
```diff
@@ -27,28 +29,7 @@ const operation = {
...
-
-  const codeSnippet = generateCodeSnippet(oas, operation, {}, {}, 'node', oasUrl, harOverride);
+  const codeSnippet = generateCodeSnippet(null, null, null, null, 'node', null, harExamples.full);
```
erunion, line 32: Made all of these arguments `null` because the only ones that should be getting picked up are the language and override.

packages/oas-to-snippet/__tests__/index.test.js
```diff
@@ -279,17 +260,61 @@ test('should not double-encode query strings', () => {
...
+      targets.forEach(target => {
+        if ('opts' in supportedLanguages[lang].httpsnippet.targets[target]) {
+          // eslint-disable-next-line jest/no-conditional-expect
+          expect(supportedLanguages[lang].httpsnippet.targets[target].opts).toStrictEqual(expect.any(Object));
+        }
+
+        if ('install' in supportedLanguages[lang].httpsnippet.targets[target]) {
+          // eslint-disable-next-line jest/no-conditional-expect
+          expect(supportedLanguages[lang].httpsnippet.targets[target].install).toStrictEqual(expect.any(String));
+        }
```
erunion, line 290: I don't really like this conditional testing, but Jest doesn't offer any "assert this value as something if it exists, if it doesn't its ok". I could do an assertion on this with JSON Schema but that'd require loading in `ajv` or some other validator. No thanks!

packages/oas-to-snippet/src/supportedLanguages.js
```diff
@@ -1,80 +1,212 @@
-const supportedLanguages = {
+module.exports = {
```
erunion, line 1: I should note that by putting all these alternative targets into these configs that we're going to be exposing them in the `APIRequest` module with some work I'm doing to support custom code samples for manual APIs. Shouldn't be a problem, but FYI regardless.

Dashron (approved):
nothing jumped out at me.

PR #1318, opened 2021-06-10
feat: moving local development up to npm@7

erunion:
| [☁️ &nbsp; CI App][demo] |
| --- |

## 🧰 What's being changed?

* [x] Moves all local development and `package.json` files over to npm@7
* [x] Updates internal package pointers to use `file:../<package name>`
  * Unclear how this'll actually interact with NPM when the package is published but hey it works locally so it might work? 
* [x] Fixed a minor bug in our `api-logs` package where our PHP SDK was being improperly named as "Php".

## 🧬 Testing

* [ ] Tests pass?
* [ ] Server run still?

[demo]: https://explorer-pr-1318.herokuapp.com/


Dashron (approved):
environment seems to be working fine, and nothing jumps out at me in the code!

